PR #26, opened 2018-01-19
Throw error if config is not valid JSON

maurelian:
##  What I did:

Made it so that an error is thrown if the config file is invalid JSON.

## Why I did it

I think it's dangerous and misleading to silently handle the error by using an empty config object.

## How to verify it

I wasn't sure how to test this, since all the tests use `linter.processStr()`, which takes a `config`, but in practice solhint.js looks for a file.


idrabenia:
Hi @maurelian ,

You changes is now available in v1.1.10.

Thank you for the contribution!
\- Ilya

PR #40, opened 2018-04-26
closes #33 closes #35

### LINKED ISSUE #33: Addition of emit keyword when logging an event
Solidity 0.4.20+ now recommends using 'emit' keyword to creating an event.  e.g. 
```emit Transfer(msg.sender, _to, _tokens);```

This creates 2 issues in solhint:
* Parsing produces something like ```line 26:21 mismatched input '(' expecting {';', '='}```
* Indentation message ```26:22  error    Expected indentation of 8 spaces but found 21  indent```

### LINKED ISSUE #35: Support latest version solc
Actually solhint doesnt support:

- event emit : `emit eventName()`
- constructor keyword in place of function name equal to contract name

yarrumretep:
Add support for new solc features: constructor syntax and emit event syntax.


mgarciap:
Thank you @yarrumretep . Somebody will review it as soon as possible

yarrumretep:
Hey guys - the emit usage warning is not great right now.  It seems hard to differentiate a warning from a function call in the old syntax (hence the introduction of emit).  This implementation gathers all the event definitions in the current file and looks to those to determine if an expression node should be an `emit`.  Of course this misses events that are defined and imported.  

Currently (unless I missed it) the linter only examines the current compilation unit (file) and does not follow imports.  Is there a better approach here?

fvictorio:
Hi @yarrumretep. Thanks for contributing and sorry for taking so long to answer.

Our grammar was taken from [this repo](https://github.com/solidityj/solidity-antlr4) and it seems to be out of sync, that's why the latest versions of Solidity are not supported. I think the best approach is to take it from there and update it.

The problem is that doing that makes a lot of tests fail. I still don't know why (I'm just starting to dig into the code), but when that is fixed, we can update this PR with the updated grammar. Does that work for you?

> Currently (unless I missed it) the linter only examines the current compilation unit (file) and does not follow imports. Is there a recommended approach to discovering the events?

I'm not sure, but I guess not? I think the parser works at the file level. Knowing that something that was inherited is an event and not a function doesn't seem to be the kind of thing a parser should know.

If that's the case, and since the compiler does warn you when you don't use emit even for inherited events, maybe we should consider just not having that rule? I'm not sure on this.

yarrumretep:
Hi @fvictorio,  Regarding the grammar - I didn't realize that you pulled the grammar from elsewhere, so I edited it in place.  It's your call if you want to integrate the changed grammar from the original source, or allow them to diverge.  Another approach would be to merge this PR and raise a new issue to merge the grammar from that source.  That way folks can get a working solhint sooner, and there is a TODO for re-sync'ing the grammar.

On the events - I guess it is fine to leave the detection of missing 'emit' keywords out.  In general I was hoping to be able to provide that additional context - but perhaps its not worth the overhead of recursively parsing all of the dependencies.  

Would you like me to remove the in-file check for emits?

yarrumretep:
@fvictorio - I've taken a stab at the following:
- integrate grammar as git submodule
- repair the incompatibilities with the updated grammar
- remove the in-file-only event emit syntax check

fvictorio:
This looks really good! I'll do a more in depth review tomorrow, but from what I skimmed I think we can already merge it. Thanks a lot!

fvictorio:
Thanks a lot for this, @yarrumretep! I just added two small commits: one fixing a typo and the other changing the error message (because for me "before 0.4.21 does not include 0.4.21, but let me know if you disagree). Awesome work :clap: :clap:

yarrumretep:
@fvictorio - thanks for merging this - and agreed on all your edits.

Are you planning on pushing a new version 1.1.11 soon?  I'd love to get the atom-solidity-linter using the latest.

fvictorio:
@yarrumretep We'll publish a new version soon. I'll let you know as soon as we do it.

fvictorio:
@yarrumretep Version 1.2.0 was just published. Thanks again for contributing!

yarrumretep:
Great stuff, @fvictorio - thank you.  And thanks for creating a great fast linter for Solidity!  Rock on!

PR #42, opened 2018-05-08
Add copyright to grammar, acknowledgement and instructions

fvictorio:


sohkai:
👏

Another thing you could do is instead of copying the grammar, install it as a git submodule, like how [solidity-parser-antlr did](https://github.com/federicobond/solidity-parser-antlr/pull/13).

fvictorio:
@sohkai That's a great idea. The js files for the parser still would need to be generated manually, but that seems OK. And it could be done when the pointed commit is updated, right?

sohkai:
Yup, I assume you'll have a script or something to update the js files... but they could be done manually whenever you update the commit you're pointing to.

fvictorio:
No, we don't have a script yet. It's just a `java -jar ...` command, but still, it would be a good idea to have one, at least as documentation.

fvictorio:
@sohkai I added a very basic script for downloading and running antlr (borrowing an idea from https://github.com/solidityj/solidity-antlr4/blob/89a1331d11d205c84ba1f542b6d5bb79ae602d1c/run-tests.sh). I also want to use a submodule for the grammar, but I think I will leave that for another PR. Thanks a lot for the suggestions :slightly_smiling_face:

fvictorio:
@mgarciap Done.

fvictorio:
ANTLR is made with Java, I don't think we can get rid of it unless we switch to another parser generator (which would probably take a lot of work) or someone ports ANTLR to another language (unlikely? it's a big and mature project).

PR #44, opened 2018-05-22
Use exact match when checking for deprecated functions

fvictorio:
Instead of using `.includes` for checking if a method is `suicide` or `sha3`, make an exact comparison with the identifier. Otherwise, functions like `get_sha3` or `suicideTimes` are interpreted as deprecated.

Note that this is still not perfect, because then you can't use `sha3` as a variable name (this seems valid, although the compiler emits a warning). But checking if the identifier is part of a function call is a little cumbersome, at least with the current grammar.


pablofullana:
LGTM 👍

matiasgaratortiz:
Looks good! 👍

PR #52, opened 2018-08-11
Change quotes documentation to match other rules

DReigada:
This makes the documentation for rules options more consistent


fvictorio:
Thanks!

PR #54, opened 2018-08-27
Add contribution formatter parameter validation

florblue:
# Description

It PR add an input parameter validation in format option.
The motivation appeared when I noticed that in the case of a invalid formatter input,  it not handle any exception. 
So, to validate the input, I decided to check it  into an array with all the valid options. Also,  it handle exception if appears an error when try to look for the required module.

To reproduce the changes:

--format option with a valid parameter (eg. -- format table [file])
--format option with an invalid parameter (eg. -- format tabe [file])

## Fixes #(issue)

There are no previous issues related.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] Existing unit tests pass locally with my changes

### DIFF
--- a/solhint.js
+++ b/solhint.js
@@ -127,14 +127,27 @@ function processPath(path) {
 }
 
 function printReports(reports, formatter) {
-    const formatterName = formatter || 'stylish';
-    if (formatterOptions.includes(formatterName)) {
-        const formatterFn = require(`eslint/lib/formatters/${formatterName}`);
+    try {
+        const formatterFn = getFormatter(formatter);
         console.log(formatterFn(reports));
         return reports;
+    } catch(ex) {
+        console.error(ex.message);
+    }
+}
+
+function getFormatter(formatter){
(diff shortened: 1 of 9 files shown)


solhint.js
```diff
@@ -5,6 +5,19 @@ const linter = require('./lib/index');
...
+    'compact',
+    'html',
+    'jslint-xml',
+    'json',
+    'junit',
+    'stylish',
+    'table',
+    'tap',
+    'unix',
+    'visualstudio'];
```
fvictorio: Since we delegate the formatting to eslint, instead of hardcoding the list of possible formatters I think a better option is to handle this in the `catch` block when the formatter is required.

The downside of this approach is that the error will be throw after the linter is executed, and that wastes user time unnecessarily. But I think it's OK for now. (A possible improvement to avoid this is to get the formatter before the linter is run.)

PR #55, opened 2018-08-28
Add --quiet option 

florblue:
# Description

It was added --quiet (-q) option that allows to report errors only, filtering out warnings reports.
Its resolves an item of Issue #5 

To do this, it was modified execMainAction(), and it was added a function that filter errors reports.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

### DIFF
--- a/solhint.js
+++ b/solhint.js
@@ -14,11 +14,9 @@ function init() {
     program
         .usage('[options] <file> [...other_files]')
         .option('-f, --formatter [name]', 'report formatter name (stylish, table, tap, unix)')
+        .option('-q, --quiet', 'report errors only - default: false')
         .description('Linter for Solidity programming language')
         .action(execMainAction);
-
-    program
-        .option('-q, --quiet', 'report errors only - default: false');
 
     program
         .command('stdin')
@@ -43,8 +41,8 @@ function execMainAction() {
     const reports =_.flatten(reportLists);
 
     if (program.quiet) {
-        // Setting int the report list errors only.
-        reports[0].reports  = getErrorResults(reports);
+        // filter the list of reports, to set errors only.
+        reports[0].reports  = reports[0].reports.filter(i => i.severity === 2);
     }
 
     printReports(reports, program.formatter);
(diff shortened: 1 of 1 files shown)


solhint.js
```diff
@@ -17,6 +17,11 @@ function init() {
         .description('Linter for Solidity programming language')
         .action(execMainAction);
 
+    program
+        .usage('[options] <file> [...other_files]')
+        .option('-q, --quiet', 'report errors only - default: false')
```
fvictorio: This is duplicating the previous block. I think you just need to add the `.option` line in that block.

solhint.js
```diff
@@ -39,6 +44,12 @@ function execMainAction() {
     const reportLists = program.args.filter(_.isString).map(processPath);
     const reports =_.flatten(reportLists);
 
+    if (program.quiet) {
+        console.log('::::Quiet mode enabled - filtering out warnings::::');
```
fvictorio: I don't think that we should log anything here.

solhint.js
```diff
@@ -113,6 +124,17 @@ function processPath(path) {
...
 }
 
+
+/**
+     * Returns results that only contains errors.
+     * @param {LintResult[]} reporter The results to filter.
+     * @returns {LintResult[]} The filtered results.
+**/
+function getErrorResults(reporter) {
+    const reporterErrors = reporter[0].reports.filter(i => i.severity === 2);
```
fvictorio: This is short and unlikely to be reused, I would do it inline.

florblue: I agree, its a good observation.

solhint.js
florblue: Perfect, I will modify it.

solhint.js
```diff
@@ -113,6 +121,17 @@ function processPath(path) {
...
 }
 
+
+/**
+     * Returns results that only contains errors.
+     * @param {LintResult[]} reporter The results to filter.
+     * @returns {LintResult[]} The filtered results.
+**/
+function getErrorResults(reporter) {
+    return reporter[0].reports.filter(i => i.severity === 2);
```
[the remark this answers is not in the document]
florblue: Ok, I get it now

Issue #56, opened 2018-08-29
[Feature request] Add max-warnings option


florblue:
Add --max-warnings [int] option in formatter option: Number of warnings to trigger nonzero exit code.

Number of warnings to trigger nonzero exit code - default: -1

The motivation comes from an option that EsLint CLI offers, and continue tha idea of issue #5.

pablofullana:
@fvictorio shall we close this one?

fvictorio:
Yep.

PR #57, opened 2018-08-30
Add --max-warnings [int] option

florblue:
# Description

It was added --max-warnings [int] option:  Number of warnings to trigger nonzero exit code.

It resolves issue #56.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update
- [x] Existing unit tests pass locally with my changes


fvictorio:
After the merge, using `-q` breaks the `--max-warnings` option. I understand `-q` as hiding the warnings in the report, not removing them altogether, so I think the two options should work together.

PR #58, opened 2018-09-05
Add new feature --ignore-path option

florblue:
# Description

It was added --ignore-path [file-name] option: allows to specify path of ignore file, and include file validation.
* This change was motivated by checking basics features that esLint gives.
* It requires a documentation update
* Existing unit tests pass locally with my changes


fvictorio:
This looks good, but there are merge conflicts.

PR #72, opened 2018-10-09
Add prettier support

mariano-aguero:
This PR adds support for prettier. It would be necessary to add tests, It will be added in the next commit.

Note: the dependency of the library `prettier-plugin-solidity` should be modified when the release comes out


mariano-aguero:
Added tests

Issue #73, opened 2018-10-09
Switch from Blacklisting to Rulesets + Whitelisting


pablofullana:
Blacklisting is quite a tricky and tedious process. Requires many iterations and isn't straightforward at all. 
Proposal is to evaluate switching this approach in favor of Rulesets + Whitelisting. Guess it that this will make Solhint setup simpler and more intuitive.

pablofullana:
@fvictorio we could discuss allowing something in the fashion of what is described in #27 as well

pablofullana:
Some code to start iterating over https://gist.github.com/pablofullana/5c1e062de0ef15e88996f3a0a17684fd

PR #76, opened 2018-11-21
Switch from Blacklisting to Whitelisting

mariano-aguero:
This is a proposal to change from blacklisting to whitelisting. See #73
We can discuss the solution raised in this pull request, ideas are welcome.

Changes made:
- Update rules to have separation of concepts and single responsibility principle in the lib folder
- Add logic to manage whitelisting
- Update test to support whitelisting
- Refactoring of test files to match rule id

Regards


package.json
```diff
@@ -16,7 +16,7 @@
   "homepage": "https://protofire.github.io/solhint/",
   "repository": "https://github.com/protofire/solhint",
   "scripts": {
-    "exec-tests": "nyc -x \"**/grammar/**\" mocha && nyc report --reporter=text-lcov | coveralls",
+    "exec-tests": "nyc -x \"**/grammar/**\" mocha --recursive \"./test/rules/**/*.js\" && nyc report --reporter=text-lcov | coveralls",
```
fvictorio: Why is this necessary?

mariano-aguero: Mocha searches for the files in the test directory. Mocha does not find the tests in the subfolder rules. I think with the recursive option it would be fine

mariano-aguero:
There is a mayor breaking change so we changed the base branch to 2.0

Regards

mariano-aguero:
> This looks good, but just to clarify: this PR does not add the rulesets feature, or am I missing something here?

It will be added in a next PR

Regards

pablofullana:
@mariano-aguero looks good. But, given:

> > This looks good, but just to clarify: this PR does not add the rulesets feature, or am I missing something here?

> It will be added in a next PR

Let's just change the PR name to match actual changes.

PR #80, opened 2018-11-27
Allow to specify the path to config file

### LINKED ISSUE #78: Allow to specify the path to config file
Sometimes you have multiple solidity projects that want to share the same solhint config. Now the only option is to copy the config file. If we allow to specify the config file location instead of hardcoding it we will be able to put that file in one place and reference it.

mariano-aguero:
Closes #78

### DIFF
--- a/lib/common/errors.js
+++ b/lib/common/errors.js
@@ -1,0 +1,15 @@
+class FileNotExistsError extends Error {
+  constructor(message, status) {
+    super(message)
+
+    this.name = this.constructor.name
+
+    Error.captureStackTrace(this, this.constructor)
+
+    this.status = status || 500
+  }
+}
+
+module.exports = {
+  FileNotExistsError
+}
--- a/lib/common/errors.js
+++ b/lib/common/errors.js
@@ -1,12 +1,7 @@
 class FileNotExistsError extends Error {
-  constructor(message, status) {
+  constructor(data) {
+    const { message } = data
     super(message)
-
-    this.name = this.constructor.name
-
-    Error.captureStackTrace(this, this.constructor)
-
-    this.status = status || 500
   }
 }
 
--- a/solhint.js
+++ b/solhint.js
@@ -144,12 +144,11 @@ const readConfig = _.memoize(() => {
   } catch (e) {
     if (e instanceof SyntaxError) {
       console.log('ERROR: Configuration file [.solhint.json] is not a valid JSON!\n')
-      process.exit(1)
     }
     if (e instanceof FileNotExistsError) {
       console.log(`ERROR: Configuration file [${program.config}] doesn't exist!\n`)
-      process.exit(1)
     }
+    process.exit(1)
   }
 
   const configExcludeFiles = _.flatten(config.excludedFiles)
--- a/solhint.js
+++ b/solhint.js
@@ -5,9 +5,11 @@ const _ = require('lodash')
 const fs = require('fs')
 const process = require('process')
 const linter = require('./lib/index')
+const packageJson = require('./package.json')
 
 function init() {
-  program.version('1.1.10')
+  const version = packageJson.version || '1.1.10'
+  program.version(version)
 
   program
     .usage('[options] <file> [...other_files]')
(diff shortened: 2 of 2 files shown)


solhint.js
```diff
@@ -16,6 +19,7 @@ function init() {
       '-w, --max-warnings [maxWarningsNumber]',
       'number of warnings to trigger nonzero exit code'
     )
+    .option('-c, --config-path [file_name]', 'file to use as your .solhint.json')
```
fvictorio: Let's use `--config` here, like eslint does.

solhint.js
```diff
@@ -5,9 +5,12 @@ const _ = require('lodash')
 const fs = require('fs')
 const process = require('process')
 const linter = require('./lib/index')
+const packageJson = require('./package.json')
+const { FileNotExistsError } = require('./lib/common/errors')
 
 function init() {
-  program.version('1.1.10')
+  const version = packageJson.version || '1.1.10'
```
fvictorio: I think we can remove the `1.1.10` fallback here.

solhint.js
```diff
@@ -128,13 +132,30 @@ const readConfig = _.memoize(() => {
...
+    }
+
     config = JSON.parse(configStr)
   } catch (e) {
     if (e instanceof SyntaxError) {
       console.log('ERROR: Configuration file [.solhint.json] is not a valid JSON!\n')
       process.exit(0)
     }
+    if (e instanceof FileNotExistsError) {
+      console.log('ERROR: Configuration file [' + program.configPath + '] doesnt exist!\n')
```
fvictorio: `doesn't`

solhint.js
```diff
@@ -128,13 +132,30 @@ const readConfig = _.memoize(() => {
...
+      process.exit(0)
```
fvictorio: This shouldn't exit with a success code.

solhint.js
```diff
@@ -128,13 +132,30 @@ const readConfig = _.memoize(() => {
   let config = {}
 
   try {
-    const configStr = fs.readFileSync('.solhint.json').toString()
+    let configStr
```
fvictorio: I think we should do:

```js
const configStr = program.configPath || '.solhint.json'
```

and then adapt the following lines accordingly.

lib/common/errors.js
```diff
@@ -0,0 +1,15 @@
+class FileNotExistsError extends Error {
+  constructor(message, status) {
+    super(message)
+
+    this.name = this.constructor.name
+
+    Error.captureStackTrace(this, this.constructor)
+
+    this.status = status || 500
```
fvictorio: This seems copied from an error for HTTP. Can we simplify it?

solhint.js
mariano-aguero: Solved

solhint.js
mariano-aguero: Solved

solhint.js
mariano-aguero: Solved!

solhint.js
mariano-aguero: Solved!

lib/common/errors.js
mariano-aguero: Solved!

solhint.js
mariano-aguero: Solved!

solhint.js
```diff
@@ -128,12 +132,23 @@ const readConfig = _.memoize(() => {
...
     config = JSON.parse(configStr)
   } catch (e) {
     if (e instanceof SyntaxError) {
       console.log('ERROR: Configuration file [.solhint.json] is not a valid JSON!\n')
-      process.exit(0)
+      process.exit(1)
+    }
+    if (e instanceof FileNotExistsError) {
+      console.log(`ERROR: Configuration file [${program.config}] doesn't exist!\n`)
+      process.exit(1)
```
pablofullana: Can't this two `process.exit(1)` lines be simplified in just one after both `if` statements?

mariano-aguero: Done

mariano-aguero:
Closes #84

PR #86, opened 2018-12-14
Add rulesets

### LINKED ISSUE #73: Switch from Blacklisting to Rulesets + Whitelisting

mariano-aguero:
Closes #73 

- Allow to extend rules to solhint through the config file (recommend, default, all)
- See custom config to use (you can create yours) https://www.npmjs.com/package/solhint-config-align-base
- Add ajv schema validator, schema and metadata to rules, and some tests
- Now you can disable a rule with the 'off' flag (eslint style)
- Enjoy it!

Example of .solhintrc.json or .solhint.json:
First install package: `npm i solhint-config-align-base`
`{
  "extends": [ "align-base" ],
  "rules": {
    "avoid-throw": "off",
    "indent": ["error", 2]
  }
}`
Or you can extend a core rule like `all`, `recommended` or `default`


Germanaz0:
### Thanks

So excited to view this PR into the project.

Regards.

lib/config/config-file.js
```diff
@@ -0,0 +1,85 @@
...
+
+const applyExtends = config => {
+  let configExtends = config.extends
+
+  // normalize into an array for easier handling
+  if (!Array.isArray(config.extends)) {
+    configExtends = [config.extends]
+  }
+
+  return configExtends.reduceRight((previousValue, parentPath) => {
```
fvictorio: Any reason to reduce from the right? This means that the rules that appear first in the array will override the ones that appear next. I think it makes more sense to do it the other way around, and that's the way eslint does it if I'm not wrong.

fvictorio:
The rules should have higher priority than the rulesets. Right now, it seems that if a rule is set by a ruleset, then you cannot override it.

PR #89, opened 2018-12-19
Have --max-warnings better reflect its name

nventuro:
Currently, a run with `--max-warnings 0` always fails, and `max-warnings 1` only succeeds if there are no warnings.

I changed the behavior and updated the description to better reflect the expected behavior.

### DIFF
--- a/solhint.js
+++ b/solhint.js
@@ -17,7 +17,7 @@ function init() {
     .option('-f, --formatter [name]', 'report formatter name (stylish, table, tap, unix)')
     .option(
       '-w, --max-warnings [maxWarningsNumber]',
-      'number of allowed warnings to not trigger nonzero exit code'
+      'number of allowed warnings'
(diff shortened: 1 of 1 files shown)


solhint.js
```diff
@@ -17,7 +17,7 @@ function init() {
     .option('-f, --formatter [name]', 'report formatter name (stylish, table, tap, unix)')
     .option(
       '-w, --max-warnings [maxWarningsNumber]',
-      'number of warnings to trigger nonzero exit code'
+      'number of allowed warnings to not trigger nonzero exit code'
```
pablofullana: Isn't something in the fashion of `number of allowed warnings` actually easier to read?

nventuro: I left the original comment about a nonzero exit code to avoid being too opinionated, but I do think your description should be enough.

PR #91, opened 2018-12-20
Added disable-previous-line.

nventuro:
Similar to `solhint-disable-next-line`, `previous-line` disables linter errors on the line preceding the comment directive.

This is very useful when disabling `no-empty-blocks` errors, since it lets you put the directive inside the (empty) block, and not before it, where there may be comments (e.g. if it is a function).

Some examples from OpenZeppelin:

```
/**
 * @dev Validation of an executed purchase. Observe state and use revert statements to undo rollback when valid
 * conditions are not met.
 * @param beneficiary Address performing the token purchase
 * @param weiAmount Value in wei involved in the purchase
 */
function _postValidatePurchase(address beneficiary, uint256 weiAmount) internal view {
    // solhint-disable-previous-line no-empty-blocks
}
```

```
/**
 * @dev This contract should only be used by extending from it, deploying it as a base contract is disabled.
 */
constructor () internal {
  // solhint-disable-previous-line no-empty-blocks
} 
```

### DIFF
--- a/test/comment-directives.js
+++ b/test/comment-directives.js
@@ -21,11 +21,24 @@ describe('Linter', () => {
       assertNoErrors(report)
     })
 
-    it('should disable only one compiler error', () => {
+    it('should disable only one compiler error on next line', () => {
       const report = linter.processStr(
         `
                 // solhint-disable-next-line
-                pragma solidity ^0.4.4; 
+                pragma solidity ^0.4.4;
+                pragma solidity 0.3.4;
+            `,
+        noIndent()
+      )
+
+      assertErrorCount(report, 1)
+    })
+
+    it('should disable only one compiler error on previous line', () => {
+      const report = linter.processStr(
+        `
+                pragma solidity ^0.4.4;
+                // solhint-disable-previous-line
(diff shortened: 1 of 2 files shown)


test/comment-directives.js
```diff
@@ -38,7 +51,7 @@ describe('Linter', () => {
       const report = linter.processStr(
         `
                 /* solhint-disable-next-line */
-                pragma solidity ^0.4.4; 
+                pragma solidity ^0.4.4;
```
nventuro: Looks like these had some empty whitespace at the end that my editor removed automatically. I can roll these changes back if that's an issue.

fvictorio: No, I'm all for removing trailing whitespace :smile:

pablofullana: No, that is absolutely fine. Thanks.

nventuro:
@fvictorio note that these tests are affected by https://github.com/protofire/solhint/pull/92: we may want to merge that one first and then fix the asserts.

nventuro:
Merged #92 into this, should be ready now.

PR #92, opened 2018-12-20
Fixed some comment directive tests.

nventuro:
The assert statements checks for `assertErrorCount(report, 1)`, but the disabled line is `pragma solidity ^0.4.4;`, which triggers a _warning_ ('compiler version must be fixed'), not an error, so the test isn't really testing whether or not `disable-next-line` works.

This makes it so that both lines have an error.


nventuro:
Done in https://github.com/protofire/solhint/pull/91/commits/99efbf9d8c8ec2ed3dd7aa0065becd970edc27f1

PR #94, opened 2018-12-20
function-order now also applies to public view/pure, not just external.

nventuro:
Added for consistency's sake. If accepted, we may also want to include `internal` and `private` here. Distinction between `view` and `pure` is also missing, is that something we want to tackle?


nventuro:
`constant` was removed in Solidity 0.5.0, so I don't think that'd be 'too harsh'.

The ordering you suggested makes sense, and is in line with the [Solidity style guide](https://solidity.readthedocs.io/en/develop/style-guide.html#order-of-functions).

fvictorio:
> `constant` was removed in Solidity 0.5.0, so I don't think that'd be 'too harsh'.

I meant forcing certain order for `view` and `pure` functions; not allowing to alternate `view` and `pure` functions, for example.

nventuro:
Ah, in that case I'd consider making it optional, yes. I don't feel too strongly about function order though, so I may not be the best person to ask about this :stuck_out_tongue:

fvictorio:
@nventuro Sorry for not merging this before. I'll make a release now so that it's available.

PR #99, opened 2019-01-04
Add plugins support

fvictorio:
Add support for writing custom rules through plugins. Check the added documentation to see how to write them.

### DIFF
--- a/docs/shareable-configs.md
+++ b/docs/shareable-configs.md
@@ -12,7 +12,7 @@ This example shows the two kind of shareable configs that you can use: the ones 
 
 ## Creating your own shareable config
 
-Shareable configs are regular npm packages. There are only two confitions:
+Shareable configs are regular npm packages. There are only two conditions:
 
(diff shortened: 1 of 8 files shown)


docs/shareable-configs.md
```diff
@@ -0,0 +1,30 @@
...
+
+```
+  "extends": ["solhint:recommended", "protofire"]
+```
+
+This example shows the two kind of shareable configs that you can use: the ones included with SolHint, that start with `solhint:`, and the ones that you can install from npm. The latter are packages that are prefixed with `solhint-config-`, so in this case the package would be installed doing `npm install solhint-config-protofire` but used as just `protofire` when adding it.
+
+## Creating your own shareable config
+
+Shareable configs are regular npm packages. There are only two confitions:
```
patitonar:
```suggestion
Shareable configs are regular npm packages. There are only two conditions:
```

Issue #111, opened 2019-04-23
Add minimum compiler version


nventuro:
Right now only `compiler-gt-0_4` is available: it'd be great if we could specify ourselves the minimum version.

PR #112, opened 2019-04-23
Add compiler-version rule.

nventuro:
Fixes https://github.com/protofire/solhint/issues/111.

`compiler-gt-0_4` could (should?) be replaced by this one, since asserting >0.4 is quite useless nowadays. I'd also bump the default value to e.g. `0.5.7`, considering [the recent bugfixes](https://blog.ethereum.org/2019/03/26/solidity-optimizer-and-abiencoderv2-bug/).

The first batch of tests are simply a copy of the `compiler-gt-0_4` ones (which are rather generic and don't actually test the version itself), but I did add a couple more specific ones at the end.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -8,7 +8,7 @@ description: List of validation rules for Solhint - Solidity security, style gui
 
 ### Options
  Default list of options are *false, "error", "warn"*. It supports by all rules.
- It provides in format 
+ It provides in format
  ```json
  {
     "rules": {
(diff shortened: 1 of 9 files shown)


fvictorio:
Thanks a lot for this!

I pushed a commit setting 0.5.7 as the default; I think it makes sense.

With respect to deprecating the older rule, I totally agree. I'm going to open an issue for that, since we don't have a mechanism for deprecating rules yet, and I don't want to block this PR.

My only question is, do you think it would make sense to allow any semver pattern, and default to `^0.5.7`?

nventuro:
>My only question is, do you think it would make sense to allow any semver pattern, and default to ^0.5.7?

Hmm, interesting. I used `^` because of the rule being called `compiler-gt`, but if we went for a semver pattern we could also deprecate `compiler-fixed` by simply using a fixed version for the pattern. I think this extra flexibility is worth it, yes - perhaps we should rename the rule?

fvictorio:
Yes, the rule should be renamed. The options that come to mind are `compiler-version`, `compiler` and `version`. Do you have any preference?

I agree with deprecating the other rule, I'll update the issue.

nventuro:
I like `compiler-version` the most, it's succinct and doesn't leave room for misinterpretation. I'll update the PR to reflect this.

fvictorio:
Thanks a lot @nventuro! We'll try to publish a new version with this rule ASAP.

PR #115, opened 2019-05-11
Fixes #114 not firing false positive on mark-callable-contracts for structs

### LINKED ISSUE #114: mark-callable-contracts rule is firing a false positive on structs
The `mark-callable-contracts` rule is being triggered against `structs`, as they are instantiated very similarly to contracts. We should update this rule so that it does not get triggered by structs. Take for instance this example:

```sol
contract ReservePool {

  struct Token {
    address tokenAddress;
    string tokenSymbol;
    uint256 currentRate;
  }

  mapping(address => Token) public acceptedTokens;
.
.
.
  function addAcceptedToken(string memory tokenSymbol, address tokenAddress, uint256 rate) public onlyAdmins {
    // solhint-disable-next-line mark-callable-contracts, solhint is confusing structs for contracts
    Token memory token = Token({ tokenAddress: tokenAddress, tokenSymbol: tokenSymbol, currentRate: rate });
    acceptedTokens[tokenAddress] = token;
  }
.
.
.
}
```

Note that `solhint-disable-next-line` had to be used for this to pass solhint as the `struct` was being confused for a contract.

habdelra:
Fixes #114

### DIFF
--- a/lib/rules/security/mark-callable-contracts.js
+++ b/lib/rules/security/mark-callable-contracts.js
@@ -25,6 +25,11 @@ class MarkCallableContractsChecker {
     this.reporter = reporter
     this.ruleId = ruleId
     this.meta = meta
+    this.structNames = []
+  }
+
+  enterStructDefinition(ctx) {
+    this.gatherStructNames(ctx)
   }
 
   exitIdentifier(ctx) {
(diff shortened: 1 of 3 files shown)


fvictorio:
Hi @habdelra, thanks a lot for the PR!

I pushed a couple of commits more (one for adding more tests, and another one for renaming a variable that had a weird name). Neither of those had to do with your changes, but they won't hurt.

My only question is about the changes in `package-lock.json`, since you didn't modify `package.json`. Can we not include those changes?

habdelra:
sure no prob

fvictorio:
Merged. Thanks @habdelra!

PR #119, opened 2019-05-16
Fixes #118 mark-callable-contracts firing false positive on constant names

### LINKED ISSUE #118: mark-callable-contracts rule is firing a false positive on constants
The mark-callable-contracts rule is being triggered against emitting events, as events are camel-cased very similarly to contracts. We should update this rule so that it does not get triggered by emitting an event. Take for instance this example:
```
contract A {
  uint8 private constant TOKEN_DECIMALS = 15;

  function b() public view returns(uint8) {
    return TOKEN_DECIMALS;
  }

}
```

habdelra:
Fixes #118


fvictorio:
Thanks!

PR #124, opened 2019-06-22
Add rules for revert reason string

### LINKED ISSUE #108: Add rules for revert reason strings
We'll be working on [revert reasons in OpenZeppelin](https://github.com/OpenZeppelin/openzeppelin-solidity/issues/1709), and it'd be great to have some sort of linter support for this. The two features I can think of are:
a) check that every `require` statement has a reason string
b) check that each reason string is at most N characters long (if it's a literal)

Do you think this is something that makes sense to have in solhint?

mariano-aguero:
Closes #108 

This is an approach to add rules for revert/require reason strings.
Take into account the following:
- Check that every require or revert statement has a reason string
- Check that each reason string is at most N characters long (if it's a literal)

Regards

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -85,4 +85,4 @@ http://solidity.readthedocs.io/en/develop/style-guide.html)
  | **function-max-lines**        | Function body contains "count" lines but allowed no more than *maxlines*. | [*\<[default](#options)\>*,&nbsp;*\<maxlines\>*] Default *maxlines* is **45**. |
  | **code-complexity**           | Function has cyclomatic complexity "current" but allowed no more than *maxcompl*. | [*\<[default](#options)\>*,&nbsp;*\<maxcompl\>*] Default *maxcompl* is **7**. |
  | **max-states-count**          | Contract has "some count" states declarations but allowed no more than *maxstates* | [*\<[default](#options)\>*,&nbsp;*\<maxstates\>*] Default *maxstates* is **15**. |
- | **reason-string**             | Ensure that revert and require statements contains error message but allowed no more than *maxLength* | [*\<[default](#options)\>*,&nbsp;*{ maxLength: 50}*] Default *maxLength* is **20**. |
+ | **reason-string**             | Ensure that revert and require statements contains error message but allowed no more than *maxLength* | [*\<[default](#options)\>*,&nbsp;*{ maxLength: 50}*] Default *maxLength* is **32**. |
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -85,3 +85,4 @@ http://solidity.readthedocs.io/en/develop/style-guide.html)
  | **function-max-lines**        | Function body contains "count" lines but allowed no more than *maxlines*. | [*\<[default](#options)\>*,&nbsp;*\<maxlines\>*] Default *maxlines* is **45**. |
  | **code-complexity**           | Function has cyclomatic complexity "current" but allowed no more than *maxcompl*. | [*\<[default](#options)\>*,&nbsp;*\<maxcompl\>*] Default *maxcompl* is **7**. |
  | **max-states-count**          | Contract has "some count" states declarations but allowed no more than *maxstates* | [*\<[default](#options)\>*,&nbsp;*\<maxstates\>*] Default *maxstates* is **15**. |
+ | **reason-string**             | Ensure that revert and require statements contains error message but allowed no more than *maxLength* | [*\<[default](#options)\>*,&nbsp;*{ maxLength: 50}*] Default *maxLength* is **20**. |
--- a/lib/rules/best-practises/reason-string.js
+++ b/lib/rules/best-practises/reason-string.js
@@ -73,7 +73,7 @@ class ReasonStringChecker extends BaseChecker {
 
   getFunctionName(ctx) {
     const parent = ctx.parentCtx
-    const { 0: name } = parent.children
+    const name = parent.children[0]
     return name.getText()
   }
 
--- a/lib/rules/best-practises/reason-string.js
+++ b/lib/rules/best-practises/reason-string.js
@@ -79,7 +79,7 @@ class ReasonStringChecker extends BaseChecker {
 
   getFunctionParameters(ctx) {
     const parent = ctx.parentCtx
-    const { 2: nodes } = parent.children
+    const nodes = parent.children[2]
     const childs = nodes.children[0].children
     const parameters = childs.filter(value => value.getText() !== ',').map(value => value.getText())
     return parameters
--- a/lib/rules/best-practises/reason-string.js
+++ b/lib/rules/best-practises/reason-string.js
@@ -1,0 +1,101 @@
+const _ = require('lodash')
+const BaseChecker = require('./../base-checker')
+
+const DEFAULT_MAX_CHARACTERS_LONG = 20
+
+const ruleId = 'reason-string'
+const meta = {
+  type: 'best-practises',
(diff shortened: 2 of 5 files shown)


test/rules/best-practises/reason-string.js
```diff
@@ -0,0 +1,125 @@
...
+    assertErrorMessage(report, 'Error message for revert is too long')
+  })
+
+  it('should not raise warning for require', () => {
+    const code = funcWith(`require(!has(role, account), "Roles: account already has role");
+        role.bearer[account] = true;role.bearer[account] = true;
+    `)
+
+    const report = linter.processStr(code, {
+      rules: { 'reason-string': ['warn', { maxLength: 50 }] }
```
fvictorio: Make `maxLength` 31 (the exact length of the reason string in the snippet) to make sure there aren't off-by-one errors.

test/rules/best-practises/reason-string.js
```diff
@@ -0,0 +1,125 @@
...
+
+    const report = linter.processStr(code, {
+      rules: { 'reason-string': ['error', { maxLength: 50 }] }
+    })
+
+    assertNoWarnings(report)
+    assertNoErrors(report)
+  })
+
+  it('should not raise warning for require with text errors', () => {
```
fvictorio: I don't understand what is being tested in these last two scenarios.

lib/rules/best-practises/reason-string.js
```diff
@@ -0,0 +1,101 @@
...
+  defaultSetup: ['warn', { maxLength: DEFAULT_MAX_CHARACTERS_LONG }],
+
+  schema: [
+    {
+      type: 'array',
+      items: [
+        {
+          properties: {
+            maxLength: {
+              type: 'string'
```
fvictorio: Shouldn't `maxLength` be of type number?

lib/rules/best-practises/reason-string.js
```diff
@@ -0,0 +1,101 @@
...
+        if (lastParameterWithoutQuotes.length > this.maxCharactersLong) {
+          this._errorMessageIsTooLong(ctx, functionName)
+        }
+      }
+    }
+  }
+
+  isReasonStringStatement(ctx) {
+    const name = this.getFunctionName(ctx)
+    return _.includes(['revert', 'require'], name)
```
fvictorio: PP: `return name === 'revert' || name === 'require'`.

lib/rules/best-practises/reason-string.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  getFunctionName(ctx) {
+    const parent = ctx.parentCtx
+    const { 0: name } = parent.children
```
fvictorio: I guess this is the same as `const name = parent.children[0]`? If that's the case, please let's use that, it's much less clever.

lib/rules/best-practises/reason-string.js
```diff
@@ -0,0 +1,101 @@
...
+    return name.getText()
+  }
+
+  getFunctionParameters(ctx) {
+    const parent = ctx.parentCtx
+    const { 2: nodes } = parent.children
```
fvictorio: Same comment as before.

lib/rules/best-practises/reason-string.js
```diff
@@ -0,0 +1,101 @@
...
+    const childs = nodes.children[0].children
```
fvictorio: `const children`

lib/rules/best-practises/reason-string.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  enterFunctionCallArguments(ctx) {
+    if (this.isReasonStringStatement(ctx)) {
+      const functionParameters = this.getFunctionParameters(ctx)
+      const functionName = this.getFunctionName(ctx)
+
+      const lastFunctionParameter = _.last(functionParameters)
+
+      const haveReasonMessage = this.haveReasonMessage(lastFunctionParameter)
```
fvictorio: Use `has` instead of `have`.

lib/rules/best-practises/reason-string.js
```diff
@@ -0,0 +1,101 @@
...
+      const haveNoMessage = functionParameters.length <= 1 || !haveReasonMessage
```
fvictorio: The logic here is a little hard to follow. I think it can be improved by doing something like this:

```
if onlyOneParameter:
  raise noReasonMessage
  return
if lastParameterTooLong
  raise reasonTooLong
```

This won't check that the last parameter is a string, but that's probably a compiler's responsibility.

docs/rules.md
```diff
@@ -85,3 +85,4 @@ http://solidity.readthedocs.io/en/develop/style-guide.html)
  | **function-max-lines**        | Function body contains "count" lines but allowed no more than *maxlines*. | [*\<[default](#options)\>*,&nbsp;*\<maxlines\>*] Default *maxlines* is **45**. |
  | **code-complexity**           | Function has cyclomatic complexity "current" but allowed no more than *maxcompl*. | [*\<[default](#options)\>*,&nbsp;*\<maxcompl\>*] Default *maxcompl* is **7**. |
  | **max-states-count**          | Contract has "some count" states declarations but allowed no more than *maxstates* | [*\<[default](#options)\>*,&nbsp;*\<maxstates\>*] Default *maxstates* is **15**. |
+ | **reason-string**             | Ensure that revert and require statements contains error message but allowed no more than *maxLength* | [*\<[default](#options)\>*,&nbsp;*{ maxLength: 50}*] Default *maxLength* is **20**. |
```
sohkai: Was there a particular reason why the default is set to 20? Solidity does the string allocation in 32-byte chunks.

fvictorio: Makes sense, let's use 32.

test/rules/best-practises/reason-string.js
mariano-aguero: Done!

lib/rules/best-practises/reason-string.js
mariano-aguero: Mmm I think should be integer, change Done!

lib/rules/best-practises/reason-string.js
mariano-aguero: Done

lib/rules/best-practises/reason-string.js
mariano-aguero: Done

lib/rules/best-practises/reason-string.js
mariano-aguero: Done

lib/rules/best-practises/reason-string.js
mariano-aguero: Done

lib/rules/best-practises/reason-string.js
mariano-aguero: Done

docs/rules.md
mariano-aguero: The default value was updated to 32

lib/rules/best-practises/reason-string.js
mariano-aguero: Logic updated

test/rules/best-practises/reason-string.js
mariano-aguero: There is no need for the last two scenarios, I remove them

pcowgill:
@fvictorio Is this good to merge? Thanks!

PR #127, opened 2019-07-09
Add config file parameter

mariano-aguero:
No issue related.

Add config file parameter as a cosmiconfig option search place

### DIFF
--- a/lib/config/config-file.js
+++ b/lib/config/config-file.js
@@ -30,7 +30,11 @@ const createEmptyConfig = () => ({
   parserOptions: {}
 })
 
-const loadConfig = () => {
+const loadConfig = configFile => {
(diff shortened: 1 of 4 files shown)


lib/config/config-file.js
```diff
@@ -47,6 +51,10 @@ const loadConfig = () => {
     ]
   }
 
+  if (configFile) {
+    cosmiconfigOptions.searchPlaces.push(configFile)
```
fvictorio: I merged this PR but then I realized this `configFile` should be added at the beginning of the `searchPlaces` array, not at the end. Otherwise, if you have a `.solhint.json` file but for some reason want to manually specify a different file, you won't be able to do it.

PR #134, opened 2019-08-07
fix #133: do not trigger mark-callable-contracts on enums

### LINKED ISSUE #133: mark-callable-contracts rule is firing a false positive on enums
The mark-callable-contracts rule is being triggered against enums (similar to #114 and #118).

Simple contract to reproduce the issue:
```
pragma solidity 0.5.9;

contract Example {

    enum Status { Initial }

    Status public status;

    constructor() public {
        status = Status.Initial;
    }
  
}
```

Produces the following warnings:
```
10:18  warning  Explicitly mark all external contracts as trusted or untrusted  mark-callable-contracts
10:25  warning  Explicitly mark all external contracts as trusted or untrusted  mark-callable-contracts
```

adetante:
I have added enum definitions and values to the `nonContractNames` list.

### DIFF
--- a/lib/rules/security/mark-callable-contracts.js
+++ b/lib/rules/security/mark-callable-contracts.js
@@ -36,6 +36,15 @@ class MarkCallableContractsChecker {
     this.gatherNonContractNames(ctx)
   }
 
+  enterEnumDefinition(ctx) {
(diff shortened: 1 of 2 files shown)


fvictorio:
Thanks!

Issue #137, opened 2019-08-13
Update examples in the "Configure linter with comments" section


fvictorio:
At the very least, the rule used for those examples is deprecated, but the whole section could probably use a revamp too.

ChrisChinchilla:
Section mentioned is this one - https://github.com/protofire/solhint/blob/master/docs/configuration.md#configure-linter-with-comments

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 100.0 DAI (100.0 USD @ $1.0/DAI)  attached to it as part of the ConsenSys fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/protofire/solhint/137/3365).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/protofire/solhint/137/3365).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $132,648.27 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

akersof:
@fvictorio please can you explain what do you mean by "deprecated", because i am checking the code and the comment rules looks ok in the documentation.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 267 years, 2 months from now.
Please review their action plans below:


**1) [akersof](https://gitcoin.co/profile/akersof) has started work.**

1) Test
2) Find issue
3) update Doc

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/protofire/solhint/137/3365).

akersof:
@fvictorio can you please explain what do you mean by "deprecated", because i am not sure to understand well if we are talking about the same thing.
`throw` is indeed deprecated in solidity but this is why this rule exists, a warning to the user who tries to use this old solidity feature.
So the example in your doc is fine.
``` /* solhint-disable avoid-throw */
  if (a > 1) {
    throw;
  }
  /* solhint-enable avoid-throw */```
By reading the code the comment directive is fine  and works well, it avoids to trigger the `"throw" is deprecated, avoid to use it.` warning.
[...]

PR #142, opened 2019-08-23
Patch for issue #137

akersof:
I rewrite the examples to use a more accurate avoid-tx-origin rule. I updated too the .solhint.json conf file given as example to change the deprecated 'false' to "off" value for removing rules, and i added a warning rule for avoid-tx-origin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,9 +137,8 @@ Or disable all validations for a group of lines:
 ```javascript
   /* solhint-disable */
   function transferTo(address to, uint amount) public {
-          require(tx.origin == owner);
-          to.call.value(amount)();
-  
+    require(tx.origin == owner);
+    to.call.value(amount)();
   }
   /* solhint-enable */
 ```
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ format:
     "extends": "solhint:default",
     "plugins": [],
     "rules": {
-      "avoid-throw": false,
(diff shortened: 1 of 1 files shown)


fvictorio:
This looks good. I think we also need to remove the deprecated rules I mentioned in the issue. Also, we should change the syntax highlighting from javascript to solidity.

akersof:
Ok, i will take on consideration your comments and and i will make this README.md more new comers friendly. thx for the feedback i am on it. Will submit it on gitcoin too.

akersof:
I think i am done. 
With my new example based on pseudo random number generation i found an issue in your code. The rule 'not-rely-on-block-hash' is not triggered anymore on newer solidity.
The reason is the 'not-rely-on-block-hash' rule checks for the old way to get the hash of a block, and this is deprecated.
Here https://github.com/protofire/solhint/blob/master/lib/rules/security/not-rely-on-block-hash.js#L25 we can see that you check for `block.blockhash` but this trigger an error now on solc because it has been deprecated in favor of `blockhash()`. i will open an issue and make PR if you want.

fvictorio:
@akersof Thanks! This looks good.

> i will open an issue and make PR if you want.

That would be awesome, thank you.

ChrisChinchilla:
@akersof and @fvictorio this also looks good to me! So if you're all happy let's get it merged and we'll get the bounty paid.

PR #148, opened 2019-08-30
Add recommended column to rules index

### LINKED ISSUE #145: Indicate which rules are recommended in the list of rules
Add a new column in the table with a check mark for the rules that are included in the `solhint:recommended` config.

touhonoob:
resolve #145


fvictorio:
Thanks!

PR #158, opened 2019-09-09
Fix 'revert' case of reason-string

kevsul:
'revert' was incorrectly being treated the same as 'require' i.e. the
reason string as the 2nd parameter. However 'revert' takes the reason
string as its first and only parameter


fvictorio:
Thank you @kevsul!

PR #160, opened 2019-09-19
fix maxWarnings when multiple reports

### LINKED ISSUE #159: maxWarnings flag doesn't work as expected when there are multiple source files
When calculating if the number of warnings exceeds the number given by the maxWarnings flag, only the first source file (report) is taken into account. Any warnings in subsequent files are not counted.

kevsul:
Fixes #159

### DIFF
--- a/solhint.js
+++ b/solhint.js
@@ -56,8 +56,7 @@ function execMainAction() {
   const reportLists = program.args.filter(_.isString).map(processPath)
   const reports = _.flatten(reportLists)
   const warningsCount = reports.reduce((acc, i) => acc + i.warningCount, 0)
-  const warningsNumberExceeded =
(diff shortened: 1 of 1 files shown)


fvictorio:
Wow, huge bug. Thanks @kevsul! Could you please run prettier so that the CI passes?

pablofullana:
Looks like something that could have been caught by a test taking into multiple reports into account. @fvictorio can we plan on adding this scenario?

kevsul:
> Wow, huge bug. Thanks @kevsul! Could you please run prettier so that the CI passes?

Done.

fvictorio:
Thanks @kevsul!

@pablofullana: I opened https://github.com/protofire/solhint/issues/161 to track this.

PR #171, opened 2019-12-13
(Feature) Use solidity parser antlr

### LINKED ISSUE #104: Use `federicobond/solidity-parser-antlr` instead of using `solidityj/solidity-antlr4` directly
This will enable us to use the AST in a simpler (higher level) way.
For example, it will let us do `node.stateMutability` instead of parsing/checking the text of each children.

fernandomg:
This PR closes #104, by:

Updating all the rules to use `solidity-parser-antlr`


lib/rules/security/not-rely-on-time.js
```diff
@@ -21,22 +21,20 @@ class NotRelyOnTimeChecker extends BaseChecker {
...
+    if (node.name === 'now') {
+      this._warn(node)
+    }
   }
 
-  _textNotAllowed(ctx, avoidedName) {
-    if (ctx.getText() === avoidedName) {
-      this._warn(ctx)
+  MemberAccess(node) {
+    if (node.memberName === 'timestamp' && node.expression.name === 'block') {
```
fvictorio:
```suggestion
    if (node.expression.name === 'block' && node.memberName === 'timestamp') {
```

lib/rules/security/not-rely-on-block-hash.js
```diff
@@ -21,14 +21,14 @@ class NotRelyOnBlockHashChecker extends BaseChecker {
     super(reporter, ruleId, meta)
   }
 
-  exitExpression(ctx) {
-    if (ctx.getText() === 'block.blockhash') {
-      this._warn(ctx)
+  MemberAccess(node) {
+    if (node.memberName === 'blockhash' && node.expression.name === 'block') {
```
fvictorio:
```suggestion
    if (node.expression.name === 'block' && node.memberName === 'blockhash') {
```

lib/rules/security/avoid-tx-origin.js
```diff
@@ -21,29 +21,10 @@ class AvoidTxOriginChecker extends BaseChecker {
...
-    if (dot && expression1.getText() === 'tx' && expression2.getText() === 'origin') {
-      this.error(ctx, 'Avoid to use tx.origin')
-    }
-  }
-
-  _nodes(ctx) {
-    if (!this._isDotExpression(ctx)) {
-      return {}
+  MemberAccess(node) {
+    if (node.memberName === 'origin' && node.expression.name === 'tx') {
```
fvictorio:
```suggestion
    if (node.expression.name === 'tx' && node.memberName === 'origin') {
```

PR #172, opened 2020-01-02
compiler-version should reject files without versions

### LINKED ISSUE #130: compiler-version should reject files without versions
If `compiler-version` is enabled, then a file like

```solidity
contract Foo {}
```

shouldn't pass that rule because there's no pragma with the solidity version.

mariano-aguero:
Closes #130 

Please see the base branch, is the `3.0`. 

#### Includes
- Added a validation in the `SourceUnit` function of the `CompilerVersionChecker` rule
- Check for the existence of the directive in the childrens
- Added a couple of tests

### DIFF
--- a/lib/rules/security/compiler-version.js
+++ b/lib/rules/security/compiler-version.js
@@ -45,11 +45,9 @@ class CompilerVersionChecker extends BaseChecker {
   }
 
   SourceUnit(node) {
-    const hasPragmaDirectiveDef = node.children.filter(
-      curItem => curItem.type === 'PragmaDirective'
-    )
+    const hasPragmaDirectiveDef = node.children.some(curItem => curItem.type === 'PragmaDirective')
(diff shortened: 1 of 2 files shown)


lib/rules/security/compiler-version.js
```diff
@@ -44,6 +44,16 @@ class CompilerVersionChecker extends BaseChecker {
     this.requirement = (config && config.getString(ruleId, DEFAULT_SEMVER)) || DEFAULT_SEMVER
   }
 
+  SourceUnit(node) {
+    const hasPragmaDirectiveDef = node.children.filter(
```
fvictorio: `node.children.some` and then `if (!hasPragmaDirective)`.

mariano-aguero: Good catch,done!

Issue #173, opened 2020-01-15
mark-callable-contracts is firing false positive for event declared after a function


mirceanis:
This is related to #116 only the false positive is fired if the event is declared **after** the function.

I added a failing test case here:
```javascript
  it('should not return error for an event defined after function', () => {
    const code = contractWith(`
      function b() public {
        emit UpdatedToken();
      }

      event UpdatedToken();
    `)

    const report = linter.processStr(code, {
      rules: { 'mark-callable-contracts': 'warn' }
    })

    assert.equal(report.warningCount, 0)
  })
```

fvictorio:
Hi @mirceanis, thanks for the report!

We are working on a new major version (3.0) and this test passes there, so this should be fixed when we publish it. We should have a beta of it soon.

Of course it would make sense to fix this in 2.0 too, but all of our bandwidth is on the new version right now. That being said, if this is a showstopper for you, let me know and I'll see if there's an easy fix.

mirceanis:
It's not a showstopper, I'm definitely glad this is already a non-issue in the future release.
Thanks for all the good work!

fvictorio:
Of course. If you want to do it, just open a PR against the `3.0` branch with it.

mirceanis:
Since it's already fixed in 3.0 I guess it's best to close it here

PR #174, opened 2020-01-16
test: no mark-callable-contracts for an event defined after a function

mirceanis:
This adds the testcase mentioned in #173 to the new release.


fvictorio:
Thanks!

PR #176, opened 2020-01-21
Add support for autofix

### LINKED ISSUE #97: Fix mode
Consider adding `--fix` so offenses are auto-fixed when possible.

Open for discussion.

fvictorio:
This PR closes #97.

Rules can now add support for fixing the issues they found. This is a very basic implementation of the autofix support, plus the fix implementation for two rules: `avoid-throw` and `avoid-sha3`.

### DIFF
--- a/lib/rules/security/avoid-throw.js
+++ b/lib/rules/security/avoid-throw.js
@@ -18,14 +18,16 @@ const meta = {
 }
 
 class AvoidThrowChecker extends BaseChecker {
-  constructor(reporter) {
+  constructor(reporter, inputSrc) {
     super(reporter, ruleId, meta)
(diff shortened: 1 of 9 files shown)


lib/rules/security/avoid-throw.js
```diff
@@ -12,17 +12,22 @@ const meta = {
...
+  fixable: true,
 
   schema: []
 }
 
 class AvoidThrowChecker extends BaseChecker {
-  constructor(reporter) {
+  constructor(reporter, inputSrc) {
     super(reporter, ruleId, meta)
+    this.inputSrc = inputSrc
```
fernandomg: This looks like it's not necessary.

PR #182, opened 2020-01-30
Validate rule schemas

fvictorio:
The schemas of the rules configurations now only represent the configuration itself (meaning, the second element of the array when setting a rule). Then, `base-checker` validates that the schema is correct. If it's not, **only a warning is emitted**, execution is not interrupted.

Lot of files to review here, but the important part is in `base-checker`.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -78,14 +78,9 @@ const meta = {
   recommended: true,
   defaultSetup: [DEFAULT_SEVERITY, DEFAULT_INDENTS],
 
-  schema: [
-    {
-      type: 'array',
-      items: [{ type: 'integer' }],
-      uniqueItems: true,
-      minItems: 2
-    }
-  ]
+  schema: {
+    type: 'integer'
+  }
 }
 
 class IndentChecker {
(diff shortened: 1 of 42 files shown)


fernandomg:
@fvictorio, it would be nice to have the `ajv` validated scheme to inform which are those _allowed values_.

```shell
node solhint.js Foo.sol
[solhint] Warning: invalid configuration for rule 'max-line-length':
  should be equal to one of the allowed values
[solhint] Warning: rule 'avoid-sha3' level is 'eror'; should be one of "error", "warn" or "off"
```

the config file:
```json
{
  "rules": {
    "avoid-sha3": "eror",
    "max-line-length": ["not-a-valid-value", 10],
    "avoid-throw": "error"
  }
}
```

what do you think?

fvictorio:
In this particular case, we can do it "by hand": checking for the first element of the array the same way it's being done when the configuration is just a string (i.e., the level).

But this won't translate to rules that use the `oneOf` validator of `ajv`. But we could use [`better-ajv-errors`](https://github.com/atlassian/better-ajv-errors) for this.

fernandomg:
I was thinking about using the `params` prop. https://ajv.js.org/#error-parameters

image

PR #188, opened 2020-02-07
Fixed link URLs

d1ll0n:
The links in this page had .html as the extension for .md files


fvictorio:
Hi @d1ll0n! Sorry for the delay, I missed your PR.

The html extensions make sense in the [page](https://protofire.github.io/solhint/), but not in the readme. But if we fix them in the readme, they stop working in the page. A lot of projects use this pattern so there must be some way to fix it properly. I will take a look into that and update this PR.

fvictorio:
Gonna try this. Thanks @d1ll0n!!

PR #221, opened 2020-05-05
Emit unused var warning for params of empty functions

### LINKED ISSUE #136: no-unused-vars not working for params of empty functions
Code:
```solidity
contract Foo {
  function foo(uint x) public returns (uint) {
  }
}
```

Config:
```json
{
  "rules": {
    "no-unused-vars": "error"
  }
}
```

No error is emitted. Notice that this does emit:
```solidity
contract Foo {
  function foo(uint x) public returns (uint) {
    return 42;
  }
}
```

fvictorio:
Closes #136.

It seems like this was done by design? But it doesn't make a lot of sense to me. It _does_ make sense for abstract functions though. Maybe this was the way it was done before that was supported? @patitonar what do you think?

Also, this is a minor breaking change, although I would argue that the difference between fixing a bug and introducing a breaking change is in the eye of the beholder.

### DIFF
--- a/lib/rules/best-practises/no-unused-vars.js
+++ b/lib/rules/best-practises/no-unused-vars.js
@@ -27,14 +27,15 @@ class NoUnusedVarsChecker extends BaseChecker {
 
   FunctionDefinition(node) {
     const funcWithoutBlock = isFuncWithoutBlock(node)
-    const emptyBlock = isEmptyBlock(node)
 
-    if (!ignoreWhen(funcWithoutBlock, emptyBlock)) {
-      VarUsageScope.activate(node)
-      node.parameters.filter(parameter => parameter.name).forEach(parameter => {
-        this._addVariable(parameter)
-      })
+    if (funcWithoutBlock) {
+      return
     }
+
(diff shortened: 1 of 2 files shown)


patitonar:
I think it makes sense for abstract functions to not emit a warning.

In solidity version `0.4.x` for the code:
```solidity
function foo(uint256 x) public returns (uint256) {}
```
The following compile warning was displayed: 
```
Unused function parameter. Remove or comment out the variable name to silence this warning.
```
One way to avoid it as suggested is to comment the variable for example as:
```solidity
function foo(uint256 /* x */) public returns (uint256) {}
```
Would a warning be emitted in this case by the rule?

For solidity versions > `0.5.x` I see the warning for `Unused function parameter` is no longer displayed by the compiler for an empty function. Should we investigate the reasoning about this change to double check this rule is correctly applied?

fvictorio:
Uhm, I think doing whatever the compiler does is a good rule of thumb here. I will close this for the time being, because the only thing I can think of is to have rules depending on the solidity version, and I don't want to go there yet.

PR #226, opened 2020-05-11
Add GitHub templates

pablofullana:
This PR adds templates for:
- Pull requests
- Bug reports
- Feature requests


fvictorio:
I personally don't like github's templates, I feel they just add unnecessary friction. They *might* be useful in repos with a lot of activity to filter out low-effort issues/PRs, but I don't think that's the case with Solhint. At the current state of the repo, between someone not opening an issue because they don't want to fill out a form, and someone forgetting to add reproduction steps, I prefer the second option.

PR #229, opened 2020-05-23
docs: mention extendable rulesets

PaulRBerg:
Clarifies what default rulesets are provided by solhint.

Please, if there are more rulesets that should be added to the list, edit my branch directly!


fvictorio:
Thanks @PaulRBerg!

PR #231, opened 2020-06-11
Add non-state-var-no-leading-underscore rule

hellwolf:
Rationale:

While we have `private-vars-leading-underscore`, sometimes people also uses underscore for non state variable. That will actually make the private state variable leading underscore almost losing its usefulness: to tell that it is a private or internal state var when seeing an underscore.

Adding a new rule `non-state-var-no-leading-underscore` specifically to forbid non state variable having lead underscore.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -10,6 +10,11 @@ module.exports = {
     return _.isNumber(configVal) && configVal > 0 ? configVal : defaultValue
   },
 
+  getBooleanByPath(path, defaultValue) {
+    const configVal = _.get(this, path)
+    return _.isBoolean(configVal) ? configVal : defaultValue
+  },
+
   getStringByPath(path, defaultValue) {
     const configVal = _.get(this, path)
     return _.isString(configVal) ? configVal : defaultValue
@@ -21,6 +26,10 @@ module.exports = {
 
   getObjectPropertyNumber(ruleName, ruleProperty, defaultValue) {
     return this.getNumberByPath(`rules["${ruleName}"][1][${ruleProperty}]`, defaultValue)
+  },
+
+  getObjectPropertyBoolean(ruleName, ruleProperty, defaultValue) {
+    return this.getBooleanByPath(`rules["${ruleName}"][1][${ruleProperty}]`, defaultValue)
   },
 
   getString(ruleName, defaultValue) {
--- a/lib/rules/index.js
+++ b/lib/rules/index.js
@@ -58,7 +58,7 @@ function coreRules(meta) {
     ...bestPractises(reporter, config, inputSrc),
     ...deprecations(reporter),
     ...miscellaneous(reporter, config, tokens),
-    ...naming(reporter),
+    ...naming(reporter, config),
     ...order(reporter, tokens),
     ...security(reporter, config, inputSrc)
   ]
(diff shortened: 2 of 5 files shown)


lib/rules/naming/non-state-var-no-leading-underscore.js
```diff
@@ -0,0 +1,32 @@
...
+const meta = {
+  type: 'naming',
+
+  docs: {
+    description: `State variable name not start with underscore.`,
+    category: 'Style Guide Rules'
+  },
+
+  isDefault: false,
+  recommended: true,
```
fvictorio, line 14: I don't think this should be in the recommended ruleset.

lib/config.js
```diff
@@ -23,6 +28,10 @@ module.exports = {
     return this.getNumberByPath(`rules["${ruleName}"][1][${ruleProperty}]`, defaultValue)
   },
 
+  getObjectPropertyBoolean(ruleName, ruleProperty, defaultValue) {
```
fvictorio, line 31: This doesn't seem to be used anywhere in the rest of the PR.

lib/rules/naming/non-state-var-no-leading-underscore.js
```diff
@@ -0,0 +1,32 @@
+const BaseChecker = require('./../base-checker')
+const naming = require('./../../common/identifier-naming')
+
+const ruleId = 'non-state-var-no-leading-underscore'
+const meta = {
+  type: 'naming',
+
+  docs: {
+    description: `State variable name not start with underscore.`,
```
fvictorio, line 9: Can you make this description a little more clear? Maybe `Only state variables can start with an underscore` or something like that?

fvictorio:
Hi @hellwolf, thanks for this and sorry for taking so long to review it. I left some comments.

Also, I'm still not sure about this rule. I don't know if this should be a new rule or an option for `private-vars-leading-underscore` (something like `strict: true`, maybe?). What do you think?

hellwolf:
Hi @fvictorio! Thanks for the review.

I have also thought about adding a option to `private-vars-leading-underscore`, but I wasn't sure because I didn't know what option name would actually make sense.

Maybe an option `applicable-to-non-state-var`, default to true to be backward compatible?

What other language linters do for this?

lib/rules/naming/non-state-var-no-leading-underscore.js
hellwolf: sure, I copy pasted from other rule and neglected this one.

lib/config.js
hellwolf: right, it was from some interim trial and errors.

fvictorio:
> What other language linters do for this kind of situations?

I googled a little but couldn't find a similar scenario

I still think a `strict` option, defaulting to false, makes sense.

@pablofullana @fernandomg  @patitonar thoughts?

hellwolf:
@fvictorio I updated code, and implemented "strict" mode for the private-vars-leading-underscore rule instead.

lib/config.js
hellwolf: Actually now it is used.

hellwolf:
Improved test cases further and fixed bugs with `SHOULD_WARN_STRICT_CASES`

fvictorio:
Hey @hellwolf, sorry for the delay (again).

I have some commits to push, but I can't because this PR was created from the master branch. Could you update the PR so that it comes from a different branch? Thanks!

hellwolf:
what branch I shoudl use?

fvictorio:
Anything but master, the name doesn't really matter.

hellwolf:
replaced by #233

PR #233, opened 2020-06-27
implement strict mode for private-vars-leading-underscore

hellwolf:
Rationale:

While we have private-vars-leading-underscore, sometimes people also uses underscore for non state variable. That will actually make the private state variable leading underscore almost losing its usefulness: to tell that it is a private or internal state var when seeing an underscore.

Adding a strict mode to `private-vars-leading-underscore` rule specifically to forbid non state variable having lead underscore.

PR #234, opened 2020-06-29
implement strict mode for private-vars-leading-underscore

hellwolf:
Rationale:

While we have private-vars-leading-underscore, sometimes people also uses underscore for non state variable. That will actually make the private state variable leading underscore almost losing its usefulness: to tell that it is a private or internal state var when seeing an underscore.

Adding a strict mode to `private-vars-leading-underscore` rule specifically to forbid non state variable having lead underscore.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -10,6 +10,11 @@ module.exports = {
     return _.isNumber(configVal) && configVal > 0 ? configVal : defaultValue
   },
 
+  getBooleanByPath(path, defaultValue) {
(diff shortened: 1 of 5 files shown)


hellwolf:
replacing #233 

@fvictorio please try again.

fvictorio:
I still can't push to your branch, but I'll create a new PR with those extra commits. Very weird, first time it happens to me.

Thanks a lot for this @hellwolf and sorry for all the problems!

PR #235, opened 2020-06-29
Minor changes after private-vars-leading-underscore PR

fvictorio:
Some changes on top of #234.

@hellwolf these are the extra changes, in case you're interested (nothing too fancy).


hellwolf:
great!

PR #238, opened 2020-07-21
Updates part of the configuration and usage

sebastinez:
Hello,

Just found this tool and am really looking forward to using it more.
Stumbled upon two things, which I guess are outdated in the README.
- The solhint.json initialization does not work with `solhint --init` instead one has to use `solhint init-config` as explained later.
- And the generated solhint.json file does not match with the one showed in the README.

Didn't want to modify the README too much, if you are interested in a rewrite I can take a look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ solhint --version
 First initialize a configuration file, if you don't have one:
 
 ```sh
-solhint --init
+solhint init-config
 ```
 
 This will create a `.solhint.json` file with the default rules enabled. Then run Solhint with one or more [Globs](https://en.wikipedia.org/wiki/Glob_(programming)) as arguments. For example, to lint all files inside `contracts` directory, you can do:
@@ -72,7 +72,6 @@ Options:
 Commands:
 
   stdin [options]                         linting of source code data provided to STDIN
-  init-config                             create configuration file for solhint
 ```
 
 ## Configuration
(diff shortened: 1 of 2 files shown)


fvictorio:
Hey @sebastinez, thanks for the PR. `--init` does work, it just wasn't published :sweat_smile: It should work now.

With respect to the json, that's not meant to show the created config, but to show how you can tweak it. Do you find that confusing? Maybe the text before it needs some rewording.

sebastinez:
Sorry @fvictorio it seems like I was a little bit early. 😄 Now it's working.

Yes I think a little rewrite regarding the initial usage process could help to make it less confusing.
I'll work on something and push it to this PR, so you can give me your feedback.

Thanks for your fast feedback!

sebastinez:
After a reread I have to say it's not that confusing any more..

Just wanted to ask, is the redundancy of `solhint --init` and `solhint init-config` something desired? Or could one of them be removed? Or do you keep it to not introduce breaking changes?

fvictorio:
I never liked `init-config` but kept it for backward compatibility. We should've removed it in 3.0.0, but I forgot to do it. I guess we'll drop it in the next major.

In the meantime, maybe it shouldn't be shown in the README, and also we should hide it from the `help` command (not sure if this is possible with commander)

> it seems like I was a little bit early

You were probably fine! I published 3.1.0 after reading your issue :sweat_smile:

sebastinez:
> In the meantime, maybe it shouldn't be shown in the README, and also we should hide it from the help command (not sure if this is possible with commander)

Done, there is a way to suppress the help output in commander, passing an options object to the command function.
```js
program
    .command('init-config', null, { noHelp:true })
    .description('create configuration file for solhint')
    .action(writeSampleConfigFile)
```

I updated the README accordingly, but I don't know why the check is failing.

Please feel free to let me know if you have any feedback.

README.md
```diff
@@ -82,23 +81,18 @@ You can use a `.solhint.json` file to configure Solhint for the whole project.
...
-
 ```json
   {
-    "extends": "solhint:recommended",
-    "plugins": [],
-    "rules": {
-      "avoid-suicide": "error",
-      "avoid-sha3": "warn"
-    }
+    "extends": "solhint:default"
```
fernandomg, line 92: so, this change can be rolled back?

sebastinez: Yes no problem, commit c840abd rolls this change back.

sebastinez:
> I've been testing your changes and I didn't happen to have the same issue with Prettier/ESLint (except that I explicitly set the arrow-parens rule and the prettier option arrowParens to "always").

You were right @fernandomg , removing the arrow parenthesis I got a passing check.

fvictorio:
Thank you @sebastinez!

PR #241, opened 2020-07-30
Support no-visibility constructors in >=0.7.0

junderw:


junderw:
Though admittedly, since pre-0.7.0 visibility was required, but post-0.7.0 it was forbidden, anyone trying to glean rules about solidity by looking at the tests will be confused when they see a default visibility test in the with-visibility fixtures... without knowing it depends on the withContract function using >=0.7.0...

But I guess that's not what the tests are for... so it seems ok.

If you want to change it I will leave this branch open to edits by maintainers.

lib/config.js
```diff
@@ -1,6 +1,22 @@
 const _ = require('lodash')
 
 module.exports = {
+  compareSemVer(a, b) {
```
fvictorio, line 4: I think you can use [semver](https://www.npmjs.com/package/semver) for this. It's already installed.

junderw: Thanks! Fixed.

junderw:
Looking forward to this. :+1:

lib/rules/deprecations/base-deprecation.js
```diff
@@ -12,8 +12,7 @@ class BaseDeprecation extends BaseChecker {
     const pragma = node.name
     const value = node.value
     if (pragma === 'solidity') {
-      this.active =
-        config.compareSemVer(value.replace(/[^0-9.]/g, ''), this.deprecationVersion()) >= 0
+      this.active = semver.gte(value.replace(/[^0-9.]/g, ''), this.deprecationVersion())
```
fvictorio, line 15: Hey @junderw, sorry for not having reviewed this yet. I will make a proper review as soon as I have some time.

For the time being, I can say that this line doesn't feel quite right. For example, this would fail if the version range is `<0.7.0`.

Actually, the more I think about this, the more I feel the deprecation mechanism doesn't make sense, because it's impossible to know which version of the compiler is actually being used. For example, let's say there's a rule that is invalid for the range `<0.7.0`, and a file has a version range `>=0.6.0`. The rule shouldn't be deprecate when using the version `0.7.0` of the compiler, but there's no way to know this from the pragma alone.

I need to analyze this more in depth. But my first instinct would be to let the user handle the problem however they see fit. For example, if someone is using 0.7.0 they should disable this rule, or add some option. Maybe instead of invalidating rules for some versions, the best approach is to give better mechanisms for users to configure the rules (e.g. disabling some rule in a directory, idk). Again, not really sure, I will think about this a little more.

junderw: Sorry for the confusion.

The BaseDeprecation change was an additional cleanup and does not change any behavior. I can remove it if necessary, I just thought it should be consistent in how it deals with semver comparisons.

Timeline:

1. I was looking for examples where this library handles semver checking and found basedeprecation.
2. The semver compare that existed was lacking, so I re-implemented it as a util and used it in my change AND basedeprecation.
3. You suggested semver package, so I changed BOTH to use semver.

junderw: re: how to handle it.

I don't know how you can get around semver gating it...

0.7.0 forbids it but 0.6.12 requires it.

No matter what error or warning we make, it will be causing an endless loop for one of those two versions of fixing compiler then fixing lint until an ignore comment is added.

fvictorio: Yes, this is totally my fault! I should've realized before that this deprecation thing doesn't really work :sweat_smile:

Do you mind rolling back all the deprecation stuff, and just leaving the updating of the rule? Sorry for all this back and forth.

junderw: Fixed.

junderw:
This should be good to merge. Any other problems that need fixing?

fvictorio:
Hey @junderw, sorry for the delay (again).

I'm still not convinced about the pragma approach, it seems too brittle (again, a `<0.7.0` would break it). I think the only option here is to add an option to the rule. I pushed a commit that does this. Can you check it out and tell me if that works for you?

junderw:
That seems reasonable, but I think we should be actively alerting the user of this specific awkward instance.

1. I updated the docs by fixing the meta info and generating the docs.
2. I change the error message when it occurs on a constructor to alert people of the new option for >=0.7.0.

I think this is the best we can do in this situation... :-/

junderw:
I have squashed the commits together.

The original commit structure at this point can be found here:

https://github.com/junderw/solhint/tree/oldPR

junderw:
(Also I have rebased on protofire master)

fvictorio:
> I updated the docs by fixing the meta info and generating the docs.

Thanks! I was going to do that and then I forgot :sweat_smile: 

> I change the error message when it occurs on a constructor to alert people of the new option for >=0.7.0.

That's a great idea, thank you for doing it!

PR #246, opened 2020-08-07
Report parse errors instead of throwing an exception

### LINKED ISSUE #245: Proper error output for parse errors
Currently parse errors cause an exception which results in `solhint` aborting and printing the Javascript stack trace. This makes it difficult to extract parse errors from the output for IDE integration for example. Ideally parse errors would show like any other hint error.

I’m happy to provide a patch.

geigerzaehler:
Report parse errors instead of throwing an exception. Fixes #245

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -13,21 +13,34 @@ function parseInput(inputStr) {
     // first we try to parse the string as we normally do
     return parser.parse(inputStr, { loc: true, range: true })
   } catch (e) {
-    try {
-      // using 'loc' may throw when inputStr is empty or only has comments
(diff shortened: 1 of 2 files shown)


fvictorio:
Thanks @geigerzaehler!

I will review and test this more in depth as soon as I can. For the time being, could you add at least one more test for an scenario with multiple parser errors (assuming this indeed can happen, I don't remember if the parser stops at the first error or not).

geigerzaehler:
> For the time being, could you add at least one more test for an scenario with multiple parser errors

Done

fvictorio:
This is awesome, thank you @geigerzaehler!

PR #247, opened 2020-08-07
Fix markdown for documentation site

geigerzaehler:
Fix the markdown syntax so that jekyll can properly parse and render it.


fvictorio:
Thank you!

PR #253, opened 2020-10-14
Get a formatting function when file from STDIN

area:
I'm guessing not many people use this functionality! With the released version, if you try to use a file from STDIN (as we do in our pre-commit git hooks, so that we're linting what's actually being committed, not what's in the file), and there are issues, `solhint` fails with:

```
TypeError: formatter is not a function
```
as `formatter` is undefined in the call to `printReports`. This PR gets a default formatter and passes it in to the function. I've not respected `--formatter` as it's not listed as an option in the `--help` for the `stdin` command.


fvictorio:
Hey @area, thanks a lot for this!

This coincided nicely with me experimenting on how to test solhint end-to-end (the only way to test things like this), so I added those tests and then I added a new one for this scenario. This made me notice that the error code was wrong (it was always 0), so I fixed that too.

There's still something missing that I didn't test yet, and that it's not included in this PR: max warnings won't work with stdin. But I think I will just open an issue for that, and use it as an excuse to refactor the `solhint.js` file a little.

area:
No worries! Thanks for getting this out so quickly.

PR #264, opened 2020-10-23
Add comprehensive-interface rule

fvictorio:
Adds a `comprehensive-interface` rule that checks that all public/external functions are overrides. This is useful to make sure that the contract's interface is extracted in a... well... interface.

cc @nventuro

### DIFF
--- a/lib/rules/miscellaneous/comprehensive-interface.js
+++ b/lib/rules/miscellaneous/comprehensive-interface.js
@@ -9,8 +9,18 @@ const meta = {
       'Check that all public or external functions are override. This is iseful to make sure that the whole API is extracted in an interface.',
     category: 'Miscellaneous',
     examples: {
-      good: require('../../../test/fixtures/miscellaneous/public-function-with-override'),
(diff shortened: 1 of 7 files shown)


lib/rules/miscellaneous/comprehensive-interface.js
```diff
@@ -0,0 +1,51 @@
...
+  FunctionDefinition(node) {
+    if (!this.isContract) {
+      return
+    }
+
+    const isPublic = node.visibility === 'public' || node.visibility === 'external'
+    const isOverride = node.override !== null
+
+    if (isPublic && !isOverride) {
+      this.error(node, 'All public methods have to be an override')
```
nventuro, line 46:
```suggestion
      this.error(node, 'All public or external methods have to be an override')
```

fvictorio:
> All public or external methods in a contract must override a definition from an interface

Makes sense, but just in case: this only checks what it says, it won't give you an error if you are overriding from a parent contract.

nventuro:
>it won't give you an error if you are overriding from a parent contract.

True, but you will also get the error at the parent at least.

PR #265, opened 2020-10-26
select files for publishing

cruzdanilo:
remove garbage and clutter from the published pack. currently, it has a lot of files that shouldn't be published, some of them are very large. for example: `e2e/node_modules/*`.

before (https://registry.npmjs.org/solhint/-/solhint-3.3.0.tgz):
Screen Shot 2020-10-26 at 23 08 29

after:
Screen Shot 2020-10-26 at 23 08 46

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,11 @@
   "bin": {
     "solhint": "solhint.js"
   },
+  "files": [
+    "/conf/",
+    "/lib/",
+    "/solhint.js"
+  ],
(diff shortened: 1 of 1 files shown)


fvictorio:
Ughh, I never noticed we didn't have a files entry, thanks for this!

I'm also going to open an issue to avoid importing things from the tests in the main code, I don't know why we are doing that.

PR #269, opened 2020-10-29
(Fix) `visibility-modifier-order` false positive with `payable` argument

### LINKED ISSUE #249: visibility-modifier-order false positive in case of payable argument and custom modifier
```solidity
contract C {
    modifier m() {
        _;
    }

    function foo(address payable addr) public payable m {
        addr;
    }
}
```

This contract triggers `Visibility modifier must be first in list of modifiers` however visibility modifier is already first in list of modifiers. Issue is reproduced only with `address payable` argument, `payable` modifier and a custom modifier.

fernandomg:
closes #249

### DIFF
--- a/lib/rules/order/visibility-modifier-order.js
+++ b/lib/rules/order/visibility-modifier-order.js
@@ -36,14 +36,28 @@ class VisibilityModifierOrderChecker extends BaseChecker {
     this.tokens = tokens
   }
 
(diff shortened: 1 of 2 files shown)


fvictorio (approved):
This implementation is awful, but it's the parser's fault. I'm going to approve this, but we should really refactor this after the parser exposes more information about the location of the visibility and state mutability modifiers.

Issue #248, opened 2020-09-18
The fallback function declared after the receive function shouldn't trigger the ordering rule


ilkecan:
[ordering](https://github.com/protofire/solhint/blob/master/docs/rules/order/ordering.md) rule shows false positive when there is an `external receive` function before the `fallback` functions. For example, for this contract:
```
contract Test {
    receive() external payable {
    }

    fallback() external payable {
    }
}
```
Solhint generates this warning:
`Function order is incorrect, fallback function can not go after external function`
which is incorrect, according to the [Style Guide](https://solidity.readthedocs.io/en/v0.7.1/style-guide.html#order-of-functions).

PR #272, opened 2020-11-19
fix: Added support for the `receive` function in `ordering`

jalextowle:
Fixes: #248


fernandomg (approved):
@jalextowle, it's looking great :100: 

Thank you so much for your contribution

PR #299, opened 2021-07-29
.find() insted .filter()[0]

spirinvladimir:


dbale-altoros:
this function is not being used

PR #322, opened 2021-12-29
make --quiet option work for all files

scott-silver:
Currently, the `--quiet` option only applies to the first file evaluated by solhint.

Say you have a directory with three contracts: `contracts/Greeter.sol`, `contracts/Greeter2.sol` and `contracts/Greeter3.sol`. All three contracts have one error and one warning.

If you run `solhint contract/**/*.sol`, you'll see all errors and warnings:

Screen Shot 2021-12-28 at 5 18 45 PM

But if you run `solhint --quiet contract/**/*.sol`, you'll see all errors for the first file (Greeter.sol) and errors+warnings for the other two files:

Screen Shot 2021-12-28 at 5 19 00 PM

Presumably, the preferred behavior is to hide warnings for all files:

Screen Shot 2021-12-28 at 5 19 48 PM

This patch hides warnings for all files when the `--quiet` flag is passed.

### DIFF
--- a/solhint.js
+++ b/solhint.js
@@ -90,7 +90,9 @@ function execMainAction() {
 
   if (program.quiet) {
     // filter the list of reports, to set errors only.
-    reports[0].reports = reports[0].reports.filter(i => i.severity === 2)
+    reports.forEach(reporter => {
+      reporter.reports = reporter.reports.filter(i => i.severity === 2)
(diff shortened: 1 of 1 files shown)


dbale-altoros:
This is addressed in 
#392 
Thanks for your contribution

Issue #281, opened 2021-02-25
virtual modifier is not supported


hellwolf:
Solidity code:

```
    //solhint-disable-next-line
    modifier onlyAuthorized(ISuperfluid host) virtual;
```

solhint output:

```
contracts/gov/SuperfluidGovernanceBase.sol
  266:53  error  Parse error: extraneous input ';' expecting {'{', 'override', 'virtual'}
  268:0   error  Parse error: extraneous input '<EOF>' expecting {'from', '}', 'using', 'struct', 'modifier', 'function', 'function', 'event', 'enum', 'address', 'mapping', 'calldata', 'var', 'bool', 'string', 'byte', 'callback', Int, Uint, Byte, Fixed, Ufixed, 'leave', 'payable', 'constructor', 'fallback', 'receive', 'receive', Identifier}
```

fvictorio:
Thanks for reporting this @hellwolf.

This depends on https://github.com/solidity-parser/parser/issues/43

patrickd-:
Ran into same issue in a Linting CI pipeline: https://github.com/patrickd-/cheatsheet.sol/runs/4956891444?check_suite_focus=true

Affected line: https://github.com/patrickd-/cheatsheet.sol/blob/6597c1d315a4bd3e940024e0d45ee12e0bf6e2a5/contracts/cheatsheet.sol#L57

---

Forked solhint and bumped @solidity-parser/parser to 0.14.0

Here: https://github.com/patrickd-/solhint/tree/patch-1

---

Changed dependency to that fork: https://github.com/patrickd-/cheatsheet.sol/commit/c16cb05fb760845719f094326c8b6eecb67e542f

Linter now succeeeds: https://github.com/patrickd-/cheatsheet.sol/runs/4980165010?check_suite_focus=true

---

Created https://github.com/protofire/solhint/pull/326

PR #326, opened 2022-01-28
Bumped @solidity-parser/parser to 0.14.0

patrickd-:
Adding virtual modifier support (https://github.com/protofire/solhint/issues/281)


k06a:
@fvictorio pls merge :)

PR #329, opened 2022-02-10
Allow low level method calls for transferring funds

### LINKED ISSUE #307: avoid-low-level-calls - false positive?
Is there a way to avoid a warning for this specific case? I believe it is currently the recommended way to transfer funds.
`(bool succeed, ) = _address.call{value: _amount}("");`

potomak:
See:

- https://solidity-by-example.org/sending-ether/
- https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now/

Closes #307


dbale-altoros:
this is continued in https://github.com/protofire/solhint/pull/394

PR #330, opened 2022-02-10
Upgrade parser

fvictorio:


potomak:
@fvictorio do you think this will resolve this error I'm getting on a block of assembly?

```
Linter: Parse error: extraneous input ',' expecting {'from', '{', '}', '(', 'error', 'for', 'function', 'address', 'calldata', 'if', 'assembly', 'return', 'revert', 'byte', 'let', '=:', 'switch', 'callback', DecimalNumber, HexNumber, HexLiteralFragment, 'break', 'continue', 'leave', 'payable', 'constructor', 'receive', Identifier, StringLiteralFragment} [undefined]
```

Code:

```
let p := j
let word := mload(add(dataPtr, p))
let h, l := byteToHex(word, hexCharsPtr)
```

The error happens on the 3rd line, but the contract compiles as expected.

potomak:
Example:

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.2;

contract Test {
    function foo() public pure {
        assembly {
            function bar() -> a, b {
                a := 1
                b := 2
            }

            let i, j := bar()
        }
    }
}
```

fvictorio:
Thank you. The parser doesn't support that syntax yet, so upgrading won't fix your problem. I created [an issue](https://github.com/solidity-parser/parser/issues/60) in the parser's repo about this.

Issue #318, opened 2021-11-26
Rule for linting console.log()?


fqlx:
Is there a rule for banning `hardhat/console.sol` and usage of `console.log`?  I often see developers forgetting to remove these debugging statements when checking in soldity code.

StoyanD:
yea, i would love it if it was added

pfmescher:
Closing this issue as already implemented

PR #335, opened 2022-02-24
feat/no-console Adding lint rule for banning console.sol and import hardhat/console.sol

fqlx:
feat/no-console Adding lint rule for banning console.sol and import hardhat/console.sol
https://github.com/protofire/solhint/issues/318


m4nuC:
Looking forward to that one!

StoyanD (approved):

PR #336, opened 2022-02-27
allow more flexible fixer utilizing current src

shir0u:
I'm trying to create a fixer for ordering rule, but cannot do that without referencing the current source.

### DIFF
--- a/solhint.js
+++ b/solhint.js
@@ -76,7 +76,7 @@ function execMainAction() {
 
       const fixes = _(report.reports)
         .filter(x => x.fix)
-        .map(x => x.fix(ruleFixer))
+        .map(x => x.fix(ruleFixer, inputSrc))
         .sort((a, b) => a.range[0] - b.range[0])
         .value()
 


dbale-altoros:
> I'm trying to create a fixer for ordering rule, but cannot do that without referencing the current source.

Can you give more context to this @shir0u  ? 
This project is gonna be supported soon

shirotech:
> > I'm trying to create a fixer for ordering rule, but cannot do that without referencing the current source.
> 
> Can you give more context to this @shir0u ? This project is gonna be supported soon

It just gives reference to original source to deal with context around reporting and fixing. One example is creating a plugin that can order/sort things, and to do that we need to know the whole original source.

shir0u:
> > I'm trying to create a fixer for ordering rule, but cannot do that without referencing the current source.
> 
> Can you give more context to this @shir0u ? This project is gonna be supported soon

It just gives reference to original source to deal with context around reporting and fixing. One example is creating a plugin that can order/sort things, and to do that we need to know the whole original source.

pfmescher (approved):
This seems harmless.. maybe it will increase memory usage a bit but otherwise I don't see why we wouldn't allow this.

fvictorio:
@shir0u the constructor of a rule receives the input source as a third parameter. So you can do this:

```
class MyRule extends BaseChecker {
  constructor(reporter, config, inputSrc) {
    super(reporter, ruleId, meta)
    this.inputSrc = inputSrc
  }
```

and then use `this.inputSrc` in your fixer (assuming it's an arrow function).

I'm not against adding this, but it seems unnecessary given that that workaround exists.

shirotech:
> @shir0u the constructor of a rule receives the input source as a third parameter. So you can do this:
> 
> ```
> class MyRule extends BaseChecker {
>   constructor(reporter, config, inputSrc) {
>     super(reporter, ruleId, meta)
>     this.inputSrc = inputSrc
>   }
> ```
> 
> and then use `this.inputSrc` in your fixer (assuming it's an arrow function).
> 
> I'm not against adding this, but it seems unnecessary given that that workaround exists.

Where are you getting these signatures from? It looks very different to https://github.com/protofire/solhint/blob/master/lib/rules/base-checker.js#L23

shir0u:
> @shir0u the constructor of a rule receives the input source as a third parameter. So you can do this:
> 
> ```
> class MyRule extends BaseChecker {
>   constructor(reporter, config, inputSrc) {
>     super(reporter, ruleId, meta)
>     this.inputSrc = inputSrc
>   }
> ```
> 
> and then use `this.inputSrc` in your fixer (assuming it's an arrow function).
> 
> I'm not against adding this, but it seems unnecessary given that that workaround exists.

Where are you getting these signatures from? It looks very different to https://github.com/protofire/solhint/blob/master/lib/rules/base-checker.js#L23

dbale-altoros:
@fvictorio so what's the status of this ?
Seems it can be dismissed ?

shir0u:
I think this can be closed, thank you all for your inputs.

Issue #239, opened 2020-07-26
TypeError: cannot ready property 'errorCount' of undefined


PaulRBerg:
## Description
Happens on Windows 10 version 1909 when running `solhint` as a local dependency installed with yarn 1.22.4.

Happens in both PowerShell and Git's Mingw64.

![Annotation 2020-07-26 163533](https://user-images.githubusercontent.com/8782666/88480324-1d14cf80-cf5e-11ea-889a-fa013eb6e43a.jpg)

## Steps to Reproduce

1. Clone https://github.com/paulrberg/solidity-template
2. Install dependencies with yarn
3. Run `yarn lint:sol`

fvictorio:
Hey @PaulRBerg, this is really weird :/ I can't reproduce it neither in Linux nor in a Windows 10 virtual machine.

PaulRBerg:
Hmm .. weird. Now it works on my end too.

I just started using Windows 10 after a long time on Mac .. maybe I got my setup wrong when I opened this issue.

fvictorio:
> I just started using Windows 10 after a long time on Mac

I'm sorry to hear that :slightly_frowning_face:

PaulRBerg:
Haha, it's temporary! My Mac had a hardware fault and I left it at the repair store.

wighawag:
I am getting the same error in my CI :

```
solhint -w 0 src/**/*.sol

/home/runner/work/sandbox-private-contracts/sandbox-private-contracts/node_modules/solhint/solhint.js:97
    if (program.maxWarnings && !reports[0].errorCount && warningsNumberExceeded) {
                                           ^

TypeError: Cannot read property 'errorCount' of undefined
    at Command.execMainAction (/home/runner/work/sandbox-private-contracts/sandbox-private-contracts/node_modules/solhint/solhint.js:97:44)
    at Command.listener (/home/runner/work/sandbox-private-contracts/sandbox-private-contracts/node_modules/solhint/node_modules/commander/index.js:315:8)
    at Command.emit (events.js:182:13)
    at Command.parseArgs (/home/runner/work/sandbox-private-contracts/sandbox-private-contracts/node_modules/solhint/node_modules/commander/index.js:656:12)
    at Command.parse (/home/runner/work/sandbox-private-contracts/sandbox-private-contracts/node_modules/solhint/node_modules/commander/index.js:474:21)
[...]

PR #351, opened 2022-08-08
fix: TypeError: Cannot read property 'errorCount' of undefined

pycckuu:
Without this patch, there are situations when the .solhintignore file
causes all inputs files to be ignored. This is a problem because the
'reports' variable ends up being an empty array, and therefore
reports[0] is undefined.

This patch solves the problem by checking that the reports array exists 
and has a length greater than 0.

Thanks to Adam Spiers (@aspiers) who originally discovered a bug.

Issue-URL: https://github.com/protofire/solhint/issues/239


aspiers:
Nice work @pycckuu :-) Please note that the correct issue URL is `https://github.com/protofire/solhint/issues/239` (or actually just #239 since we're already in this repo!) That bytemeta site is just a vampire attack clone of GitHub, which does not deserve being linked to ;-)

dbeal-eth:
@aspiers can we get this merged? we ran into this issue today and its stopping us from using `--max-warnings=0` on our CI for some reason

aspiers:
I would love that but I don't maintain this repository.

dbale-altoros (approved):

dbale-altoros (approved):

PR #352, opened 2022-08-16
fix: bug directories with .sol in the name treated as files

pycckuu:
Without this patch, the path, for example,
artifacts/@openzeppelin/contracts/token/ERC20/ERC20.sol/ERC20.dbg.js
will be treated as a file
artifacts/@openzeppelin/contracts/token/ERC20/ERC20.sol
on MacOS system. As a result, npm module fs throws the error:
'Error: EISDIR: illegal operation on a directory'

This patch solves the problem by checking if the path is a file and not
a directory.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -61,7 +61,7 @@ function processFile(file, config) {
 function processPath(path, config) {
   const ignoreFilter = ignore().add(config.excludedFiles)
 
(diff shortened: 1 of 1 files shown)


dbale-altoros:
This is continued in #388 with the corresponding E2E tests

PR #377, opened 2023-01-06
Fix of ordering rule for CustomErrorDefinition

pahor167:
Currently when smart contract contains Custom errors such as:

error MyError();

Ordering rule fails since it throws "Unrecognized contract part, please report this issue".

### DIFF
--- a/lib/rules/order/ordering.js
+++ b/lib/rules/order/ordering.js
@@ -137,6 +137,10 @@ function contractPartOrder(node) {
     return [30, 'event definition']
   }
 
+  if (node.type === 'CustomErrorDefinition') {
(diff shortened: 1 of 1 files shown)


ym73 (approved):
Support Custom errors in contract (otherwise Ordering rule fails)

Credits to [pahor167](https://github.com/pahor167)

juanpcapurro (commented):
Consider adding a unit test that fails without the addtion to ordering.js and succeeds with it, ideally in a commit before 958a1f9 so it's easier to check

dbale-altoros:
Continue in #393 with added unit tests for Custom Errors on ordering-rule

PR #381, opened 2023-01-18
Linter fixes

juanpcapurro:
Please consider configuring github merge to do a linear/semi-linear history on merge preserving history so:

- when reading a blame/log, automated linter changes can be told apart from ones made by hoomans
- contributor credit is registered in git history as well as github repository

PR #382, opened 2023-01-18
Added: linter fixes

dbale-altoros:
This PR only contains Linter fixes and some minor linter exception to avoid `import/extensions` errors when executing linter into the repo:
Exceptions added manually
[ // eslint-disable-next-line ]
lib/rules/security/mark-callable-contracts.js ==> Line 7
test/rules/security/mark-callable-contracts.js ==> Line 18

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,13 +1,13 @@
 module.exports = {
   parserOptions: {
-    ecmaVersion: 8
+    ecmaVersion: 8,
   },
   env: {
     browser: false,
     node: true,
     commonjs: true,
     es6: true,
-    mocha: true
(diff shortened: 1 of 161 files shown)


fvictorio:
@dbale-altoros can you split this PR into multiple commits?

One of those commits should just include the `prettier --write` changes, since those are easy to verify by just running prettier locally. The other commits should have the rest of the changes. That way a reviewer can ignore the first commit (the prettier changes) and review the rest.

fvictorio:
I wrote that comment before seeing @juanpcapurro's description in #381; those are also good reasons to use multiple commits here.

PR #388, opened 2023-01-25
fix: directories with .sol in the name path treated as files

dbale-altoros:
This PR adds E2E tests for the pr #352 
Credits to @pycckuu


e2e/04-dotSol-on-path/contracts/ERC20.sol/Foo.sol.sol
```diff
@@ -0,0 +1,4 @@
+pragma solidity >=0.5.0;
```
fvictorio, line 1: `Foo.sol.sol`?

e2e/test.js
```diff
@@ -111,4 +111,16 @@ describe('e2e', function() {
       expect(stdout.trim()).to.contain('Code contains empty blocks')
     })
   })
+
+  describe('.sol on path', function() {
+    useFixture('04-dotSol-on-path')
+
+    it('should success', function() {
```
fvictorio, line 118:
```suggestion
    it('should handle directory names that end with .sol', function() {
```

CHANGELOG.md
```diff
@@ -1,12 +1,24 @@
+## [3.3.9] - TBD
```
fvictorio, line 1: This change doesn't seem to belong to this PR.

e2e/test.js
```diff
@@ -111,4 +111,16 @@ describe('e2e', function() {
...
+      const { code, stdout } = shell.exec('solhint contracts/**/*.sol')
+
+      expect(code).to.equal(0)
+
+      console.log(stdout)
```
fvictorio, line 123: This console.log doesn't seem necessary.

lib/index.js
```diff
@@ -61,7 +61,7 @@ function processFile(file, config) {
 function processPath(path, config) {
   const ignoreFilter = ignore({ allowRelativePaths: true }).add(config.excludedFiles)
 
-  const allFiles = glob.sync(path, {})
+  const allFiles = glob.sync(path, {}).filter((f) => fs.lstatSync(f).isFile())
```
fvictorio, line 64: `glob` has a `nodir` option that we can use instead of doing this.

fvictorio (changes requested):

CHANGELOG.md
dbale-altoros: I corrected the old changelog and added this pr as well
The point is to have it up to date and update it on every change after this one

dbale-altoros:
@fvictorio 
I corrected all issues you requested, some of them belong to the original coder.
The changelog is modified because I added this fix and also corrected miss information of previous version
Do you think it can be approved so we can move on ?
Thanks!

fvictorio (approved):

PR #389, opened 2023-01-25
Fix doc generation

juanpcapurro:
while creating a new rule I found that the script responsible for generating the documentation was broken (well, not the script itself, but the metadata supplied to it) I fixed it and added a CI step to ensure that at least it doesn't crash.

I tried to get CI to also check that the latest version of the docs are commited to source control (with `git diff docs --exit-code`) but versions for files were different between CI and my machine. So it's out of scope for this PR :upside_down_face: but I can create an issue to implement that later if there's consensus

### DIFF
--- a/.github/workflows/E2E.yml
+++ b/.github/workflows/E2E.yml
@@ -20,10 +20,14 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - uses: actions/checkout@v2
+      - name: install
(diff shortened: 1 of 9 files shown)


dbale-altoros:
@juanpcapurro 
This is great. I noticed before the generator had an error and it couldn't get the job done.
Will review with Franco

dbale-altoros (approved):

PR #391, opened 2023-01-26
fix: no-console rule update

dbale-altoros:
There was an error in the no-console rule preventing to invoke any method in a variable including the 'console' string.
This update fix that. Tests included.

### DIFF
--- a/lib/rules/best-practises/no-console.js
+++ b/lib/rules/best-practises/no-console.js
@@ -17,11 +17,11 @@ const meta = {
         },
         {
           description: 'No hardhat/console.sol import statements',
-          code: 'import "hardhat/console.sol"' // eslint-disable-line
+          code: 'import "hardhat/console.sol"', // eslint-disable-line
         },
         {
           description: 'No forge-std console.sol & console2.sol import statements',
-          code: 'import "forge-std/consoleN.sol"' // eslint-disable-line
+          code: 'import "forge-std/consoleN.sol"', // eslint-disable-line
         },
       ],
     },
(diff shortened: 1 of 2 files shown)


pfmescher (approved):

pfmescher (approved):

lib/rules/best-practises/no-console.js
```diff
@@ -17,11 +17,11 @@ const meta = {
         },
         {
           description: 'No hardhat/console.sol import statements',
-          code: 'import "hardhat/console.sol"' // eslint-disable-line
+          code: 'import "hardhat/console.sol"', // eslint-disable-line
         },
         {
           description: 'No forge-std console.sol & console2.sol import statements',
-          code: 'import "forge-std/consoleN.sol"' // eslint-disable-line
+          code: 'import "forge-std/consoleN.sol"', // eslint-disable-line
```
fvictorio, line 24: I didn't notice this in the previous review, but this `// eslint-disable-line` comments shouldn't be used unless there is a good reason. It seems that in the previous implementation they were used to fix the warning about the missing comma, and now they were kept after the proper fix was done.

lib/rules/best-practises/no-console.js
```diff
@@ -41,24 +41,31 @@ class NoConsoleLogChecker extends BaseChecker {
...
+      node.expression.expression.type === 'Identifier' &&
+      (node.expression.expression.name === 'console' ||
+        node.expression.expression.name === 'console2') &&
+      node.expression.memberName.startsWith('log')
+    )
   }
 
   isAnImport(node) {
     return (
       node.type === 'ImportDirective' &&
```
fvictorio, line 65: I missed this in the previous review, but this check doesn't make sense anymore. Since you are using the `ImportDirective` callback, this check will always be true.

I think I would just inline this logic in the `ImportDirective` callback, this indirection is not necessary.

fvictorio (changes requested):

lib/rules/best-practises/no-console.js
dbale-altoros: Sorry I don't understand what you mean.

As I understood: 
ImportDirective is executed if/when the code has an import... so that is calling the isAnImport() function to check the import path

FunctionCall is executed if/when a call to a function is being done... like console.log, so that is why the expression makes sense

I did delete line 42 which was useless

Please post your suggestion or approve the pr if you consider we can move past this. Thanks!

lib/rules/best-practises/no-console.js
dbale-altoros: done

lib/rules/best-practises/no-console.js
fvictorio: The `ImportDirective` callback is called when an import node is visited, so checking if it's of type 'ImportDirective' is redundant. I pushed a commit fixing it.

fvictorio (approved):

lib/rules/best-practises/no-console.js
dbale-altoros: ohhh sorry
I thought you were talking about these checks
```
node.expression.expression.name === 'console' ||
node.expression.expression.name === 'console2') &&
node.expression.memberName.startsWith('log')
```

that's why did not make sense for me...

Yes, it was useless to check if that was an importDirective since it is being called by that node

PR #393, opened 2023-01-27
fix: ordering-rule custom error support

dbale-altoros:
This is the #377 with unit tests addition to check the behavior

credits to @pahor167

### DIFF
--- a/test/rules/order/ordering.js
+++ b/test/rules/order/ordering.js
@@ -145,10 +145,114 @@ describe('Linter - ordering', () => {
     assert.ok(report.messages[0].message.includes('Function order is incorrect'))
   })
 
-  it('should not raise error when external const goes before public ', () => {
+  it('should not raise error when external function goes before public ', () => {
     const code = contractWith(`
                 function a() external view {}
                 function b() public {}
+            `)
+
+    const report = linter.processStr(code, {
+      rules: { ordering: 'error' },
+    })
+
+    assert.equal(report.errorCount, 0)
+  })
+
+  it('should raise an error when custom error is after external function', () => {
+    const code = contractWith(`
+                function a() external view {}
+                error Unauthorized();
+                function b() public {}
+            `)
+
+    const report = linter.processStr(code, {
+      rules: { ordering: 'error' },
+    })
+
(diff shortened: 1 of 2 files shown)


pfmescher (approved):

fvictorio:
Custom errors can also be at the file-level (that is, outside a contract) and this PR doesn't handle that.

dbale-altoros:
> Custom errors can also be at the file-level (that is, outside a contract) and this PR doesn't handle that.

@fvictorio 
good catch!
does this rule includes free functions and constant variables (file level) ?

fvictorio:
I don't remember, check the implementation.

dbale-altoros:
Added support for:

At File Level: 
- custom errors
- constants
- free functions

At Contract Level: 
- custom errors

test/rules/order/ordering.js
```diff
@@ -185,4 +289,142 @@ describe('Linter - ordering', () => {
...
+      // SPDX-License-Identifier: MIT
+      pragma solidity ^0.8.0;
+      import "@openzeppelin/contracts/ownership/Ownable.sol";
+      error Unauthorized();
+      enum MyEnum { A, B }
+      struct OneNiceStruct { uint256 a; uint256 b; }            
+      uint256 constant oneNiceConstant = 1;
+      function freeFunction() pure returns (uint256) {
+        return 1;
+      }
```
fvictorio, line 421: @frangio sorry for the ping, but do you have an opinion on this order for file-level nodes? The new supported elements are custom errors, constants and functions.

I feel like I would swap constants and errors here, so that the order is:
- Constants
- Enum/Structs
- Custom errors
- Functions
- Contracts

Bikeshedding though.

frangio: No opinion here to be honest.

fvictorio (approved):

test/rules/order/ordering.js
dbale-altoros: @fvictorio I feel the order you proposed is a better one
I changed the weight so it can match what you suggested
Please approve when you have time

fvictorio (approved):

Issue #30, opened 2018-02-04
Comments should not be counted in the function-max-lines rule


weijiekoh:
May I suggest ignoring comments when counting the number of lines in a function to check whether it exceeds function-max-lines?

i.e. the following should only count as 2 lines of code:
```
uint foo = 0; // comment
// comment
Event(bar);
```

This will prevent solhint from giving function-max-lines errors for functions which exceed 50 lines but are mainly comprised of comments. IMO, it's good practice to comment smart contract code and this shouldn't be penalised by the tool. Thanks!

dbale-altoros:
This is implemented in this PR
https://github.com/protofire/solhint/pull/395
Waiting for review

fvictorio:
If that PR is waiting for review, it shouldn't be in a draft state and the description should be updated.

PR #395, opened 2023-01-31
DRAFT - fix: function-max-lines considering comments

dbale-altoros:
This pr is to solve issue #30 
function-max-lines was considering comments and empty lines 
Now this behavior is changed. 

- empty lines
- one line comments
- and multiline comments 

Are not considered

STILL IN PROGRESS DO NOT MERGE


dbale-altoros:
@fvictorio 
It will be good if you can see this PR
I changed the core behavior and I'm not sure if there is a case where a statement can be considered as such
Or why the empty lines were considered as an actual line

Issue #339, opened 2022-04-21
Ignore "free functions" when checking for visibility. 


kaspergff:
When creating a free function (a  function outside of a contract) I get the following error: Linter: **Explicitly mark visibility in function [func-visibility**

But as explained [here](https://blog.soliditylang.org/2020/09/02/solidity-0.7.1-release-announcement/#:%7E:text=A%20free%20function%20behaves%20like,and%20in%20this%20case%2C%20msg.) 

> "Free functions are always internal functions and are meant to replace internal library functions and their very special behaviour."

. Therefore they don't need a visibility keyword.

paulrberg:
Can confirmed that this is still occurring on the latest version (v3.3.7).

It's a shame because I would like not to disable the `func-visibility` rule.

Issue #310, opened 2021-09-14
Solhint tags free functions for visibility?


grapevinegizmos:
Can't get rid of linter message on free function:

If i put
`function roleValue(Role role) pure returns(bytes32) {...}`

before the contract body, I get an linter warning `Explicity mark visibility in function [func-visiblity] [7,1]` which of course is unavoidable since you can't mark visibility on a free function.

dbale-altoros:
Same Issue ?
[#339](https://github.com/protofire/solhint/issues/339)

Issue #276, opened 2021-01-06
solidity >= 0.7.4 - free functions visibility 


gitpusha:
Solidity greater than 0.7.4 allows you to have file-level function definitions. 

Solidity forces you not to give them a `visibility modifier` because they are always `pure` by default. 

`solhint` still complains about visibility modifier specification for such solidity function definition files.

dbale-altoros:
this is the same as this one right ?
[#310](https://github.com/protofire/solhint/issues/310)

dbale-altoros:
solved here
https://github.com/protofire/solhint/pull/396

PR #396, opened 2023-02-01
fix: skip free functions on func-visibility

dbale-altoros:
This pr fixed the warning of func-visivility rule when using solidity free functions 
I think this one solves:
https://github.com/protofire/solhint/issues/276
https://github.com/protofire/solhint/issues/310
https://github.com/protofire/solhint/issues/339

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -59,7 +59,7 @@ title:       "Rule Index of Solhint"
 | [avoid-tx-origin](./rules/security/avoid-tx-origin.md)                 | Avoid to use tx.origin.                                                  | ✔️          |
 | [check-send-result](./rules/security/check-send-result.md)             | Check result of "send" call.                                             | ✔️          |
 | [compiler-version](./rules/security/compiler-version.md)               | Compiler version must satisfy a semver requirement.                      | ✔️          |
-| [func-visibility](./rules/security/func-visibility.md)                 | Explicitly mark visibility in function.                                  | ✔️          |
+| [func-visibility](./rules/security/func-visibility.md)                 | Explicitly mark visibility in function (free functions are skipped).     | ✔️          |
 | [mark-callable-contracts](./rules/security/mark-callable-contracts.md) | Explicitly mark all external contracts as trusted or untrusted.          |             |
 | [multiple-sends](./rules/security/multiple-sends.md)                   | Avoid multiple calls of "send" method in single transaction.             | ✔️          |
 | [no-complex-fallback](./rules/security/no-complex-fallback.md)         | Fallback function must be simple.                                        | ✔️          |
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -59,7 +59,7 @@ title:       "Rule Index of Solhint"
 | [avoid-tx-origin](./rules/security/avoid-tx-origin.md)                 | Avoid to use tx.origin.                                                  | ✔️          |
 | [check-send-result](./rules/security/check-send-result.md)             | Check result of "send" call.                                             | ✔️          |
 | [compiler-version](./rules/security/compiler-version.md)               | Compiler version must satisfy a semver requirement.                      | ✔️          |
-| [func-visibility](./rules/security/func-visibility.md)                 | Explicitly mark visibility in function (free functions are skipped).     | ✔️          |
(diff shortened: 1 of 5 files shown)


docs/rules.md
```diff
@@ -59,7 +59,7 @@ title:       "Rule Index of Solhint"
 | [avoid-tx-origin](./rules/security/avoid-tx-origin.md)                 | Avoid to use tx.origin.                                                  | ✔️          |
 | [check-send-result](./rules/security/check-send-result.md)             | Check result of "send" call.                                             | ✔️          |
 | [compiler-version](./rules/security/compiler-version.md)               | Compiler version must satisfy a semver requirement.                      | ✔️          |
-| [func-visibility](./rules/security/func-visibility.md)                 | Explicitly mark visibility in function.                                  | ✔️          |
+| [func-visibility](./rules/security/func-visibility.md)                 | Explicitly mark visibility in function (free functions are skipped).     | ✔️          |
```
fvictorio, line 62: Unnecessary clarification

docs/rules/security/func-visibility.md
```diff
@@ -12,7 +12,7 @@ title:       "func-visibility | Solhint"
 
 
 ## Description
-Explicitly mark visibility in function.
+Explicitly mark visibility in function (free functions are skipped).
```
fvictorio, line 15: Same.

lib/rules/security/func-visibility.js
```diff
@@ -10,7 +10,7 @@ const meta = {
   type: 'security',
 
   docs: {
-    description: `Explicitly mark visibility in function.`,
+    description: `Explicitly mark visibility in function (free functions are skipped).`,
```
fvictorio, line 13: Same.

test/fixtures/security/contracts-with-free-functions.js
```diff
@@ -0,0 +1,176 @@
+// all contracts are made with two func-visibility errors
+
+module.exports = [
```
fvictorio, line 3: I would prefer these tests to be simpler and to do one thing each. For example, a simple test with just a free function that doesn't throw errors; another test that has a free function and a contract function without visibility; etc. I realize that this means having them in different files and being more explicit in the test file (importing each fixture in a different test because it produces a different number of warnings), but it's worth it.

lib/rules/security/func-visibility.js
```diff
@@ -61,18 +63,44 @@ class FuncVisibilityChecker extends BaseChecker {
       config && config.getObjectPropertyBoolean(ruleId, 'ignoreConstructors', false)
   }
 
+  SourceUnit(node) {
```
fvictorio, line 66: A better way to do this is to set a variable when a contract definition is entered, and only apply the rule in those cases. See [this](https://github.com/protofire/solhint/blob/129251621156093557f4aeedba240bb5182f1733/lib/rules/miscellaneous/comprehensive-interface.js#L39-L45) for an example of how to do that.

fvictorio (changes requested):

docs/rules.md
dbale-altoros: removed

docs/rules/security/func-visibility.md
dbale-altoros: removed

test/fixtures/security/contracts-with-free-functions.js
dbale-altoros: done

lib/rules/security/func-visibility.js
dbale-altoros: done

lib/rules/security/func-visibility.js
```diff
@@ -61,18 +61,28 @@ class FuncVisibilityChecker extends BaseChecker {
       config && config.getObjectPropertyBoolean(ruleId, 'ignoreConstructors', false)
   }
 
+  ContractDefinition(node) {
+    this.isContract = node.kind === 'contract'
```
fvictorio, line 65: This will always be true, because the `ContractDefinition` callback is called when a contract node is visited.

```suggestion
    this.isContract = true
```

dbale-altoros: done

fvictorio (approved):

Issue #250, opened 2020-09-23
no-unused-vars false positive in case of address payable argument without name


ZumZoom:
```solidity
contract C {
    function foo(address payable /* addr */) public view {
        this;
    }
}
```

This contract triggers `Variable "payable" is unused` however payable is not a variable.

dbale-altoros:
should be closed with this PR
https://github.com/protofire/solhint/pull/399

PR #399, opened 2023-02-07
fix: no-unused-vars false positive for unnamed payable parameters

dbale-altoros:
This is a PR for the issue:
https://github.com/protofire/solhint/issues/250

Seems like it is fixed with this exception
Unit tests added


fvictorio:
This seems to be a problem in the parser, but this fix should be enough for now.

fvictorio (approved):

PR #406, opened 2023-02-17
[ new release ] updated changelog and version

dbale-altoros:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "solhint",
-  "version": "3.3.8",
```
pfmescher, line 3: Based on the amount of changes I'd say this should increase the minor and not just the patch version. I think it should be 3.4.0

pfmescher (approved):

PR #417, opened 2023-02-28
[ New Rule ] No unused imports

juanpcapurro:
I've modified both [openzeppelin-contracts](https://github.com/OpenZeppelin/openzeppelin-contracts) and [exactly](https://github.com/exactly/protocol) to manually add this rule and check for false positives, couldn't find any in the current implementation. However, I'm concerned I might be forgetting to register some usage regarding inline assembly or some other less-used feature, so feedback is of course appreciated on that.

as always, see git history for details.

### DIFF
--- a/lib/rules/best-practises/index.js
+++ b/lib/rules/best-practises/index.js
@@ -8,6 +8,7 @@ const PayableFallbackChecker = require('./payable-fallback')
 const ReasonStringChecker = require('./reason-string')
 const NoConsoleLogChecker = require('./no-console')
 const NoGlobalImportsChecker = require('./no-global-import')
+const NoUnusedImportsChecker = require('./no-unused-import')
 
 module.exports = function checkers(reporter, config, inputSrc) {
   return [
@@ -21,5 +22,6 @@ module.exports = function checkers(reporter, config, inputSrc) {
     new ReasonStringChecker(reporter, config),
     new NoConsoleLogChecker(reporter),
     new NoGlobalImportsChecker(reporter),
+    new NoUnusedImportsChecker(reporter),
   ]
 }
--- a/lib/rules/best-practises/no-unused-import.js
+++ b/lib/rules/best-practises/no-unused-import.js
@@ -52,9 +52,13 @@ class NoUnusedImportsChecker extends BaseChecker {
   }
 
   ImportDirective(node) {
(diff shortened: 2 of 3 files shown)


dbale-altoros:
@juanpcapurro This is a cool rule! Thanks for that.
This is marked as draft... not sure if it's not finished yet, or you put it that way because it depends on the previous PR

juanpcapurro:
> @juanpcapurro This is a cool rule! Thanks for that. This is marked as draft... not sure if it's not finished yet, or you put it that way because it depends on the previous PR

Hi Diego, this PR is not in a mergable state sadly, since it still has many false positives yet. so far I've found the implementation errors out when an imported name is used as:

- [ ] type for
    - [ ] variable inside a function
    - [ ] struct definition
- [ ] value for
    - [ ] revert statement
    - [ ] accessing one if its members

and also has the following errors:

- [ ] reports original name when aliased name is not used

so a full review isn't in order yet. However, I'd of course greatly appreciate feedback on how I'm approaching the fix, if I'm organizing the code right, those kinds of things.

Small chores, such as not repeating myself for the different errors that have the same message, I'll probably fix later when I have the entire feature figured out though.

dbale-altoros:
Ok, no worries... You made such great contributions here... I cannot say other thing than thank you.
I don't know how much time I will be given to approach next tasks on solhint... as soon as I have some spare time I'll try to give you some feedback (if any)

juanpcapurro:
@fvictorio @dbale-altoros  :sparkles: should be ready for review now :sparkles:

dbale-altoros (approved):
I tested it and seems pretty tight
@fvictorio what do you think ?

juanpcapurro:
> Great to see this rule being added, any idea on when the next release will be?

This rule is already released in [solhint-community](https://github.com/solhint-community/solhint-community/) 3.5.1, where a few devs and myself are continuing development

dbale-altoros:
@juanpcapurro Como estas ?
Te consulto algo.
Veo que esta regla asi como esta codeada tiene varios falsos positivos que fueron arreglados en la version de community
Te molestaria si tomo los fixes de ese repo para esta regla asi la dejamos bien aca tambien ? 

Gracias y disculpa la molestia

dbale-altoros:
@dbale-altoros check

Issue #242, opened 2020-07-31
no-empty-blocks false positive in inherited constructor


quezak:
Hello! I'm migrating from ethlint and found the same issue that I [reported there](https://github.com/duaraghav8/Ethlint/issues/264) some time ago :)

In an inherited contract, the constructor often just calls the base constructor, and its block is empty. This should be an exception from the `no-empty-block` rule, but now it causes a `warning    Code contains empty block`.

Copied from the [official example](https://solidity.readthedocs.io/en/v0.7.0/contracts.html#arguments-for-base-constructors).
```solidity
contract Base {
    uint x;
    constructor(uint _x) { x = _x; }
}

// or through a "modifier" of the derived constructor.
contract Derived2 is Base {
    constructor(uint _y) Base(_y * _y) {}
}
```

fvictorio:
Yes, this makes a lot of sense! Should be easy to implement. I can give you some pointers if you feel like giving this a shot. If not, I'll try to fix it soon.

emretepedev:
same issue here.

Janther:
Ethlint did address this issue duaraghav8/Ethlint/issues/264

johnnyshankman:
Same here as well.

sebastianst:
@fvictorio any progress here?

dbale-altoros:
Close with 
https://github.com/protofire/solhint/pull/418

PR #418, opened 2023-03-01
support `ignoreConstructors` option for `no-empty-blocks`

hackartists:
# A new example
You can set `ignoreConstructors` simlar with `func-visibility` rule.

```json
{
  "rules": {
    "no-empty-blocks": ["warn", {ignoreConstructors: true}]
  }
}
```

# Related Issue
#242

### DIFF
--- a/lib/rules/best-practises/no-empty-blocks.js
+++ b/lib/rules/best-practises/no-empty-blocks.js
@@ -27,12 +27,8 @@ class NoEmptyBlocksChecker extends BaseChecker {
     this._validateContractPartsCount(node)
   }
 
-  FunctionDefinition(node) {
-    node.body.shouldIgnore = node.isConstructor && node.modifiers.length > 0
-  }
-
   Block(node) {
-    if (node.shouldIgnore) {
+    if (node.parent.isConstructor && node.parent.modifiers.length > 0) {
       return
     }
 
--- a/lib/rules/best-practises/no-empty-blocks.js
+++ b/lib/rules/best-practises/no-empty-blocks.js
@@ -1,9 +1,5 @@
 const BaseChecker = require('../base-checker')
 const { isFallbackFunction } = require('../../common/ast-types')
-const { isArray } = require('lodash')
-const DEFAULT_IGNORE_CONSTRUCTORS = false
-const DEFAULT_OPTION = { ignoreConstructors: DEFAULT_IGNORE_CONSTRUCTORS }
-const config = require('../../config')
 
 const ruleId = 'no-empty-blocks'
 const meta = {
@@ -12,37 +8,18 @@ const meta = {
   docs: {
     description: 'Code contains empty block.',
     category: 'Best Practise Rules',
-    options: [
-      {
-        description:
-          'A JSON object with a single property "ignoreConstructors" specifying if the rule should ignore constructors.',
-        default: JSON.stringify(DEFAULT_OPTION),
-      },
-    ],
   },
 
   isDefault: false,
   recommended: true,
-  defaultSetup: ['warn', DEFAULT_OPTION],
+  defaultSetup: 'warn',
 
-  schema: {
-    type: 'object',
-    properties: {
-      ignoreConstructors: {
-        type: 'boolean',
-      },
-    },
-  },
+  schema: null,
 }
 
 class NoEmptyBlocksChecker extends BaseChecker {
   constructor(reporter) {
     super(reporter, ruleId, meta)
-
-    if (reporter) {
-      const conf = config.from({ rules: reporter.config })
-      this.ignoreConstructors = conf.getObjectPropertyBoolean(ruleId, 'ignoreConstructors', false)
-    }
   }
 
   ContractDefinition(node) {
@@ -51,13 +28,14 @@ class NoEmptyBlocksChecker extends BaseChecker {
   }
 
   FunctionDefinition(node) {
-    node.body.isConstructor = node.isConstructor
+    node.body.shouldIgnore = node.isConstructor && node.modifiers.length > 0
   }
 
   Block(node) {
-    if (node.isConstructor && this.ignoreConstructors) {
+    if (node.shouldIgnore) {
       return
     }
+
     const isFallbackFunctionBlock = isFallbackFunction(node.parent)
     if (isFallbackFunctionBlock) {
       // ignore empty blocks in fallback functions
--- a/lib/rules/best-practises/no-empty-blocks.js
+++ b/lib/rules/best-practises/no-empty-blocks.js
@@ -1,5 +1,9 @@
 const BaseChecker = require('../base-checker')
 const { isFallbackFunction } = require('../../common/ast-types')
+const { isArray } = require('lodash')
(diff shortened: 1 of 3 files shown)


juanpcapurro (commented):
I like this implementation ✨

However there's a thing I'd like to discuss:

the following is valid with the `ignoreConstructors` option

```
constuctor() {}
```

however the code doesn't do anything useful, and IMO it'd be good practice to
get rid of it.

the case in which this rule (without the option herein defined) is annoying is
where you have a constructor with a member initialization list:

```
constructor(uint a) Base(a) {}
```

since the linter complains about the empty code block but you can't get rid of
the constructor definition, since it's necessary to call the base constructor.

Therefore I propose to change the default behaviour of the rule so it *always*
ignores constructors with both a member initialization list and an empty body
and also *always* raises an error on an empty constructor without said
initialization

quick patch on top of `584479629` to show what I mean:

```
diff --git a/lib/rules/best-practises/no-empty-blocks.js b/lib/rules/best-practises/no-empty-blocks.js
index ba14fa5..1cd88dc 100644
--- a/lib/rules/best-practises/no-empty-blocks.js
+++ b/lib/rules/best-practises/no-empty-blocks.js
@@ -54,0 +55 @@ class NoEmptyBlocksChecker extends BaseChecker {
+    node.body.modifiers = node.modifiers
@@ -58 +59 @@ class NoEmptyBlocksChecker extends BaseChecker {
-    if (node.isConstructor && this.ignoreConstructors) {
+    if (node.isConstructor && node.modifiers.length > 0) {
diff --git a/test/rules/best-practises/no-empty-blocks.js b/test/rules/best-practises/no-empty-blocks.js
index 83ae325..5727461 100644
--- a/test/rules/best-practises/no-empty-blocks.js
+++ b/test/rules/best-practises/no-empty-blocks.js
@@ -29,0 +30 @@ describe('Linter - no-empty-blocks', () => {
+    contractWith('constructor (uint256 a) Base(a) {  }'),
@@ -117,10 +117,0 @@ describe('Linter - no-empty-blocks', () => {
-  it('should not raise error for constructor when ignoreConstructors is set to true', () => {
-    const code = contractWith(`constructor () {}`)
-
-    const report = linter.processStr(code, {
-      rules: { 'no-empty-blocks': ['warn', { ignoreConstructors: true }] },
-    })
-
-    assertNoWarnings(report)
-  })
-
```

what do y'all think?

hackartists:
@juanpcapurro Thank you for your comments. I agree your comments because I mostly use empty constructor only for inherited contract. 
However, it might has inteded empty block of constructor. It is why the solidity doesn't allow to skip consructor except for `abstract contract`. For example, you can often see as below code in javascript. If the class don't need to do something when instantiation.
```
const instance = new Class();
```

In this case, javascript allow us to skip writing `constructor`, but the solidity.
As the above reason, it is the reason why I declare `ignoreConstructors`.  

Your comments do works for some, so how about adding an optional setting as below
```
{
      rules: { 'no-empty-blocks': ['warn', { ignoreConstructors: true }] },
}
```
or
```
{
      rules: { 'no-empty-blocks': ['warn', { ignoreConstructors: { mustHasModifiers: 1 } }] },
}
```

How do you think about the above optional setting?

hackartists:
@juanpcapurro oh, thank you, good to me. And I will reflect your feedback.

hackartists:
@juanpcapurro I've pushed the commit reflecting your feedback.

juanpcapurro:
@dbale-altoros Could you enable CI for this PR?

lib/rules/best-practises/no-empty-blocks.js
```diff
@@ -27,7 +27,15 @@ class NoEmptyBlocksChecker extends BaseChecker {
     this._validateContractPartsCount(node)
   }
 
+  FunctionDefinition(node) {
+    node.body.shouldIgnore = node.isConstructor && node.modifiers.length > 0
```
juanpcapurro, line 31: Keep in mind the AST is shared for all visitors, so this property will exist in the Blocks that are bodies of a function in all rule definitions that are executed after this one.

Therefore, an effort should be made not to pollute the AST with information not globally relevant.

My personally preferred approach would be to add the `isConstructor` and `modifiers` fields of the `FunctionDefinition` node to its `body`, and then assert `isConstructors && modifiers.length > 0` in the listener for `Block` nodes.

Having those two names in the `Block` node makes it more clear what they are if they show up in the execution of another rule.

However, I'm not sure that it's a good practise to modify the AST that's gonna be used by other rules, and I'd like the opinion of @fvictorio on this topic. (I grepped a bit and couldn't find many usages of the `node` variable as an lvalue)

If it's problematic to modify the AST in this way, then you'd have to store the information of what blocks to ignore in the `ComplexityScore` class. One such way could be to have an object whose keys are a string that looks like ``${range[0]-range[1]}``, and it's value is wether to ignore the block or not.

juanpcapurro (commented):
the blocking issue currently is wether to modify the AST or not, and the redundant testcases are secondary

lib/rules/best-practises/no-empty-blocks.js
fvictorio: The way this is done in other rules is by adding a field to the checker instance. So:

```
this.shouldIgnore = ...
```

Agree that modifying the node is not a good idea.

hackartists: @fvictorio @juanpcapurro I agree what you've mentioned. Howeve whether to be constructor could be validated in `FunctionDefinition`. Therefore, I needed to recognize if its block is passed by constructor or not. Do you have any other good idea to pass it to the block.

test/rules/best-practises/no-empty-blocks.js
```diff
@@ -113,6 +115,26 @@ describe('Linter - no-empty-blocks', () => {
     assertNoWarnings(report)
   })
 
+  it('should raise error for empty block of [constructor]', () => {
```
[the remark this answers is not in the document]
hackartists: You are right. They were redundant so I've removed them.

test/rules/best-practises/no-empty-blocks.js
```diff
@@ -113,6 +115,26 @@ describe('Linter - no-empty-blocks', () => {
...
+    const code = contractWith(`constructor () {}`)
+
+    const report = linter.processStr(code, {
+      rules: { 'no-empty-blocks': 'warn' },
+    })
+
+    assertWarnsCount(report, 1)
+  })
+
+  it('should not raise error for empty block of inherited [constructor] having modifiers', () => {
```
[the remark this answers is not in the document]
hackartists: Basically, it was removed. But I denote the syntax keywards by brackets as the way others do

lib/rules/best-practises/no-empty-blocks.js
hackartists: @fvictorio Thank for good to me. I've change it to use node.parent and checked it working properly.

dbale-altoros (approved):

dbale-altoros:
@hackartists thanks for this contribution!!
It is a great addition
Well done!

Issue #206, opened 2020-04-01
Feature request: Allow to set the path for the plugins


juanfranblanco:
Hi I hope you are all well,

Vscode is not able to load the plugins as these will not be packaged with the extension, to workaround this it will be ideal if a path could be passed to load the plugins together with the list of names.

https://github.com/protofire/solhint/blob/32ec781509c9b76db325d0c2db91c8c478441d6a/lib/rules/index.js#L53

This is how it is handled in prettier.

Suggestion: The solhint.json file could include a path or allow to set this by the ide by setting the path to be the same as the current project node_modules. 

Reference: https://github.com/juanfranblanco/vscode-solidity/issues/169

fvictorio:
Hi @juanfranblanco, sorry for not responding before. I'm not 100% sure I understand what would work for you. In the linked issue you suggest having a flag similar to prettier's `--plugin-search-dir`, is that correct? If we implement that, would that solve the issue for you?

juanfranblanco:
Don't worry, responding issues is really hard and takes time. 

You are right that is how it works with prettier. So that could be the usual magic of letting the user set it on the solhint file, in the extension itself or if nothing is set, use the the default node_modules in the workspace.

This is how I have integrated with prettier if it helps: 
https://github.com/juanfranblanco/vscode-solidity/blob/master/src/formatter/prettierFormatter.ts#L17-L18

fvictorio:
I see, makes sense. Will try that approach then, thanks :+1: 

I guess you are using solhint as an executable binary, right? In version 3.0 we tried to also properly export it as a "library", so maybe you want to try this. That's not documented yet, but it's just:

```
const { processStr } = require('solhint')

processStr(source, config, filename)
```

Both config and filename are optional. In this case, I guess you'd pass the `pluginSearchDirs` options there.

fvictorio:
Oh, and we should document what is returned by that function :man_facepalming: I'll try to do it when I have some time, but for the time being you can probably figure it out by inspecting it :sweat_smile: Let me know if you have any questions about that.

juanfranblanco:
Solium is used as follows: https://github.com/juanfranblanco/vscode-solidity/blob/master/src/linter/solium.ts it was contributed by yourselves.

fvictorio:
This is solhint, not solium :smile:

juanfranblanco:
Oh my!! Sorry :)

juanfranblanco:
[...]

PR #424, opened 2023-03-24
ISSUE #206 - Plugins can be in any path

lhemerly:
Change loadPlugin function at index.js to receive the plugin path.

### DIFF
--- a/lib/rules/index.js
+++ b/lib/rules/index.js
@@ -67,15 +67,13 @@ function coreRules(meta) {
   ]
 }
 
-function loadPlugin(pluginName, { reporter, config, inputSrc, fileName }) {
+function loadPlugin(pluginPath, { reporter, config, inputSrc, fileName }) {
   let plugins
   try {
-    plugins = require(`solhint-plugin-${pluginName}`)
+    plugins = require(`${pluginPath}`)
   } catch (e) {
     console.error(
-      chalk.red(
-        `[solhint] Error: Could not load solhint-plugin-${pluginName}, make sure it's installed.`
-      )
+      chalk.red(`[solhint] Error: Could not load ${pluginPath}, make sure it's installed.`)
     )
     process.exit(1)
   }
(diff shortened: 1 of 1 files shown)


dbale-altoros (approved):

dbale-altoros:
@lhemerly
Thanks for this one

frangio:
Doesn't this break existing configs which specify `'foo'` in order to load `solhint-plugin-foo`?

dbale-altoros:
@frangio thanks for your input. It is always appreciate.
It might break.
We wanted to test the whole thing as an executable and evaluate how hard it might be to fix and implement this.
That is why we pushed into develop instead of master.

lhemerly:
It will break for existing config files passing the plugin name, as it needs the plugin complete path to work now. One option is to parse the string the loadPlugin method gets from the config file to use the old loading method when it starts with solhint-plugin and to use the new one (full path) when it doesn't.

cruzdanilo:
this is a breaking change. please, bump the major version following [semver](https://semver.org/#spec-item-8):

> Major version X (X.y.z | X > 0) MUST be incremented if any backward incompatible changes are introduced to the public API.

frangio:
For what it's worth, the original reason why this PR was opened was resolved [in a different way](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4132/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519).

I don't think this feature is worth a major version bump, which has a lot of downstream effects.

If there is interest in adding the native ability to load plugins from a file system path, I would suggest detecting if a plugin starts with `./` or `/` and using that as a path, otherwise using the previous behavior, to remain backwards compatible.

lhemerly:
I believe allowing to pass the plugin path, as stated in this issue, allows for easier integration of custom rules and aligns with the method of major linters. I believe @frangio

PR #428, opened 2023-04-02
Check for github action updates monthly

jlosito:
This will use dependabot to check the github actions that this project uses for any updates and submit pull requests with version bumps in order to try to keep everything up-to-date.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"


dbale-altoros:
@jlosito sorry for the delay

We something like that enabled on the repo and it generates the PRs
Like this:

image

What's the difference of what you are suggesting ?

jlosito:
@dbale-altoros looks to be the equivalent.

dbale-altoros:
@jlosito Ok, I will close this one...
Feel free to add any ideas or thoughts
We really appreciate community involvement and we're trying our best to achieve the best product 

Stay tuned for the next version coming by next week

PR #431, opened 2023-05-12
fix: --max-warnings unusable

treasure-t1dev:
The --max-warnings flag is never considered due to a mis-referenced parameter


dbale-altoros (approved):

dbale-altoros:
Added in this one
https://github.com/protofire/solhint/pull/439

PR #432, opened 2023-05-17
Add json formatter

ArturLukianov:
Add support for JSON formatter.

JSON formatter is very useful in cases where solhint is a part of tooling pipeline. It makes easy to implement additional processing on top of solhint results, without need to parse "table" or "unix" formatters output


dbale-altoros:
Added with additional functionality in this PR
https://github.com/protofire/solhint/pull/440

PR #453, opened 2023-07-14
Allow `$` in casing-related rules

### LINKED ISSUE #452: The `var-name-mixedcase` rule triggers when using `$` as a variable name
## Description

Given Solidity allows using a `$` as a valid variable identifier, consider the following example:

```solidity
contract A {
    function example() external pure returns (uint256) {
        uint256 $ = 1;
        return $;
    }
}
```

The variable `$` is marked as `Linter: Variable name must be in mixedCase [var-name-mixedcase]`.

## Proposed solution

The library should not throw an error when the variable includes a `$`

ernestognw:
Fixes #452

I acknowledge the issue requires a bit more discussion but I'm opening this PR in case there's no push back for this rule change.

Alternatively, we can pass it as part of the rule configuration, similar to:

```js
{
  "rules": {
    "var-name-mixedcase": ["error", { ignore$: true }]
  }
}
```

### DIFF
--- a/test/rules/naming/const-name-snakecase.js
+++ b/test/rules/naming/const-name-snakecase.js
@@ -54,4 +54,23 @@ describe('Linter - const-name-snakecase', () => {
     assert.equal(report.errorCount, 1)
     assert.ok(report.messages[0].message.includes('SNAKE_CASE'))
   })
+
+  describe('with $ character', () => {
+    const WITH_$ = {
+      $: contractWith('uint32 private constant $ = 10;'),
+      'starting with $': contractWith('uint32 private constant $THE_CONSTANT = 10;'),
+      'containing a $': contractWith('uint32 private constant THE_$_CONSTANT = 10;'),
+      'ending with $': contractWith('uint32 private constant THE_CONSTANT$ = 10;'),
+    }
+
+    for (const [key, code] of Object.entries(WITH_$)) {
+      it(`should not raise event name error for events ${key}`, () => {
+        const report = linter.processStr(code, {
+          rules: { 'const-name-snakecase': 'error' },
+        })
+
+        assert.equal(report.errorCount, 0)
+      })
+    }
+  })
 })
--- a/test/rules/naming/event-name-camelcase.js
+++ b/test/rules/naming/event-name-camelcase.js
@@ -13,4 +13,33 @@ describe('Linter - event-name-camelcase', () => {
     assert.equal(report.errorCount, 1)
     assert.ok(report.messages[0].message.includes('CamelCase'))
   })
+
+  it('should not raise event name error for event in camelCase', () => {
+    const code = contractWith('event Event1(uint a);')
+
+    const report = linter.processStr(code, {
+      rules: { 'event-name-camelcase': 'error' },
+    })
+
+    assert.equal(report.errorCount, 0)
+  })
+
+  describe('with $ character', () => {
+    const WITH_$ = {
+      $: contractWith('event $(uint a);'),
+      'starting with $': contractWith('event $Event1(uint a);'),
+      'containing a $': contractWith('event Eve$nt1(uint a);'),
+      'ending with $': contractWith('event Event1$(uint a);'),
+    }
+
+    for (const [key, code] of Object.entries(WITH_$)) {
+      it(`should not raise event name error for events ${key}`, () => {
+        const report = linter.processStr(code, {
+          rules: { 'contract-name-camelcase': 'error' },
+        })
+
+        assert.equal(report.errorCount, 0)
+      })
+    }
+  })
 })
--- a/test/rules/naming/func-param-name-mixedcase.js
+++ b/test/rules/naming/func-param-name-mixedcase.js
@@ -3,25 +3,87 @@ const linter = require('../../../lib/index')
 const contractWith = require('../../common/contract-builder').contractWith
 
 describe('Linter - func-param-name-mixedcase', () => {
-  it('should raise incorrect func param name error', () => {
-    const code = contractWith('function funcName (uint A) public {}')
+  describe('in functions', () => {
+    it('should raise incorrect func param name error', () => {
+      const code = contractWith('function funcName (uint A) public {}')
 
-    const report = linter.processStr(code, {
-      rules: { 'func-param-name-mixedcase': 'error' },
+      const report = linter.processStr(code, {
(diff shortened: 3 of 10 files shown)


dbale-altoros:
@ernestognw Thanks a lot for your input
The thing with this is that the isMixedCase() function is being called by:
lib/rules/naming/func-param-name-mixedcase.js
lib/rules/naming/func-name-mixedcase.js
lib/rules/naming/modifier-name-mixedcase.js
lib/rules/naming/var-name-mixedcase.js

So we need to make sure it doesn't break any other ruleset

Also, as I understand in the lib/common/identifier-naming.js there are 
isCamelCase()
isNotCamelCase()
isUpperSnakeCase()
isNotUpperSnakeCase() 

and (I'm not sure) but I think `$` can also be present in function names, constants and parameter so, I'm thinking that if we add this, we should also add in those as well (?) just thinking out loud

ernestognw:
Hey @dbale-altoros,

Yes, I think generally the mixed case should not fail if there's a `$` at the beginning of its definition. Note that the rule already allows `$` in between. So perhaps the discussion is if we should allow it at the beginning (or standalone).

I would say that allowing only standalone `$` would be the less breaking change but the regex is already half of the way by allowing it after the first character.

I can update the other rules if you think it's the way to move forward

dbale-altoros:
yes, the main thing is the leading $

I'm thinking if this really worth the hassle... 
I understand that solhint shouldn't see that as a warning/error since it's allowed. That's for sure and we appreciate your input. Really!

If you are up to check in the other rules and add tests to make sure this won't break, we can move forward... if not, we can wait on this fix since doesn't seem to be a high priority

What do you think ?

ernestognw:
Thanks also for your prompt response! Sure I think I can take a look at other rules and add tests. 

Just confirming, should I look only for rules using the following functions?

> isCamelCase()
> isNotCamelCase()
> isUpperSnakeCase()
> isNotUpperSnakeCase()

dbale-altoros:
yes, I think so
because in your pr the pattern being changed is only in that file with those functions

dbale-altoros:
I also changed the base from master into develop
please put it as draft until it is ready

thanks a lot for your contribution and time... we appreciate!!!

ernestognw:
Thanks for the guidance @dbale-altoros, I just marked this as ready for review since I already added tests for the rules affected. Should be enough but please let me know if there's something I'm missing.

dbale-altoros:
of course... this week or next it will be reviewed by team
Thanks a lot for your contribution

dbale-altoros:
@ernestognw sorry this is taking so long
we made a new rule called `immutable-vars-naming` changing some behavior on `const-name-snakecase` and `var-name-mixed-case`

So this pr has a few conflicts, I don't want to mess it up
I'll be reviewing it today or tomorrow

test/rules/naming/const-name-snakecase.js
```diff
@@ -64,8 +64,25 @@ describe('Linter - const-name-snakecase', () => {
...
+  describe('with $ character', () => {
+    const WITH_$ = {
+      $: contractWith('uint32 private constant $ = 10;'),
+      'starting with $': contractWith('uint32 private constant $THE_CONSTANT = 10;'),
+      'containing a $': contractWith('uint32 private constant THE_$_CONSTANT = 10;'),
+      'ending with $': contractWith('uint32 private constant THE_CONSTANT$ = 10;'),
+    }
+
+    for (const [key, code] of Object.entries(WITH_$)) {
+      it(`should not raise event name error for events ${key}`, () => {
```
dbale-altoros, line 77: seems this description is not ok, right ?
Don't worry I'll fix... just wanted to make sure

test/rules/naming/event-name-camelcase.js
```diff
@@ -13,4 +13,33 @@ describe('Linter - event-name-camelcase', () => {
...
+      $: contractWith('event $(uint a);'),
+      'starting with $': contractWith('event $Event1(uint a);'),
+      'containing a $': contractWith('event Eve$nt1(uint a);'),
+      'ending with $': contractWith('event Event1$(uint a);'),
+    }
+
+    for (const [key, code] of Object.entries(WITH_$)) {
+      it(`should not raise event name error for events ${key}`, () => {
+        const report = linter.processStr(code, {
+          rules: { 'contract-name-camelcase': 'error' },
```
dbale-altoros, line 38: this one should be event-name-camelcase, right ? (I fix)

test/rules/naming/func-param-name-mixedcase.js
```diff
@@ -3,25 +3,87 @@ const linter = require('../../../lib/index')
...
+    describe('with $ character', () => {
+      const WITH_$ = {
+        $: contractWith('event Event1(uint $);'),
+        'starting with $': contractWith('event Event1(uint $param);'),
+        'containing a $': contractWith('event Event1(uint pa$ram);'),
+        'ending with $': contractWith('event Event1(uint param$);'),
+      }
+
+      for (const [key, code] of Object.entries(WITH_$)) {
+        it(`should not raise func name error for functions ${key}`, () => {
```
dbale-altoros, line 79: for Event Parameters

test/rules/naming/modifier-name-mixedcase.js
```diff
@@ -23,4 +23,23 @@ describe('Linter - modifier-name-mixedcase', () => {
...
+  describe('with $ character', () => {
+    const WITH_$ = {
+      $: contractWith('modifier $(address a) { }'),
+      'starting with $': contractWith('modifier $ownedBy(address a) { }'),
+      'containing a $': contractWith('modifier owned$By(address a) { }'),
+      'ending with $': contractWith('modifier ownedBy$(address a) { }'),
+    }
+
+    for (const [key, code] of Object.entries(WITH_$)) {
+      it(`should not raise func name error for functions ${key}`, () => {
```
dbale-altoros, line 36: ==> for modifiers
(wrong description on IT statement)

dbale-altoros:
I will close this PR
I opened this one
https://github.com/protofire/solhint/pull/465

With these modifications and fixes

test/rules/naming/modifier-name-mixedcase.js
ernestognw:
```suggestion
      it(`should not raise func name error for modifiers ${key}`, () => {
```

Issue #461, opened 2023-07-30
dynamic loading of rules breaks bundlers


llllvvuu:
If you follow this:

https://github.com/protofire/solhint/blob/d182f16712ed472277ea6d044f407b12937b02fe/lib/config/config-file.js#L8-L11

https://github.com/protofire/solhint/blob/d182f16712ed472277ea6d044f407b12937b02fe/lib/load-rules.js#L9-L12

These are all dynamic references to files that won't exist at runtime if running a bundle with `eslint` bundled (e.g. using `esbuild`, `rollup`, or `webpack`).

In comparison, here is how ESLint works: https://github.com/eslint/eslint/blob/main/packages/js/src/configs/eslint-recommended.js

They do not locate the `isDefault` or `isRecommended` information next to the rule:
https://github.com/eslint/eslint/blob/main/lib/rules/no-empty.js

vs `solhint` which does:
https://github.com/protofire/solhint/blob/master/lib/rules/naming/func-name-mixedcase.js

IMO the `eslint` way is preferred also for the reason that you can quickly glance at and edit the rulesets, and is cleaner and simpler

llllvvuu:
I can put a PR for this real quick, just by dumping the resolved `enabledRules` into `jq --sort-keys`.

dbale-altoros:
@llllvvuu great work !!
The team will be reviewing this

Thanks a lot for the contribution!!!

PR #462, opened 2023-07-30
fix: remove runtime dependency on `load-rules`

llllvvuu:
dynamic loading/requiring of files breaks bundlers as those files will not be available at runtime of bundled code (#461)

i accomplish this by adding a script, `npm run generate-rulesets`, which generates a static version of the rulesets

i also add prepublish and precommit hook to make sure docs and rulesets are up-to-date if anything in the rules directory has changed


dbale-altoros (commented):
why is this file `solhint.js` empty ?

package.json
```diff
@@ -20,7 +20,10 @@
     "test:coverage": "npm run exec-tests",
     "test": "mocha --recursive",
     "lint": "eslint .",
-    "docs": "node scripts/generate-rule-docs.js"
+    "generate-rulesets": "node scripts/generate-rulesets.js && prettier --write conf/rulesets",
+    "docs": "node scripts/generate-rule-docs.js",
+    "prepare": "husky install",
```
dbale-altoros, line 25: Sorry if it is a dumb question, but
are the 'prepare' and 'preoublishOnly' needed ?

llllvvuu: It's not strictly necessary, but it makes docs and ruleset JSON automatically updated when you commit or publish

llllvvuu:
`solhint.js` was a mistake, removed

package.json
llllvvuu: Oh, I see what you're asking.

Those are [`npm` lifecycle scripts](https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts). `npm` will run `prepare` when you run `npm install`, that makes sure that `git` is synced with `husky`.

`npm` will run `prepublishOnly` when you run `npm publish`, that makes sure that `npm` package has up to date rulesets.

Issue #360, opened 2022-10-31
Enforce Function Calls with Named Parameters


oneleo:
Is it possible to be able to enforce [Function Calls with Named Parameters](https://docs.soliditylang.org/en/latest/control-structures.html#function-calls-with-named-parameters) when function containing six or more arguments?

Passes:
```solidity
function doFunc( in1, in2) external view returns () {}
function doFunc({ parm1: in1, parm2: in2, parm3: in3, parm4: in4, parm5: in5, parm6: in6 }) external view returns () {}
```

Warns or Error:
```solidity
function doFunc( in1, in2, in3, in4, in5, in6) external view returns () {}
function doFunc({ parm1: in1, parm2: in2 }) external view returns () {}
```

Thank you very much!

PaulRBerg:
This would be awesome!

dbale-altoros:
Added in https://github.com/protofire/solhint/pull/468

oneleo:
Thanks!

PR #468, opened 2023-08-03
New Rule: func-named-parameters

dbale-altoros:
requested in 
https://github.com/protofire/solhint/issues/360


fiction-code (approved):

pcaversaccio:
Hey, I would like to report a small bug here. The rule is enforced as well on the built-in functions like `keccak256`, `abi.encode`, `abi.encodePacked` for which you can't use named parameters.

PR #494, opened 2023-08-19
Fix generate docs links

vladyan18:
The old version of the script did not correctly generate links when using Windows OS due to differences in the path format.


dbale-altoros:
@vladyan18 
Great catch!
Thanks for making this PR
We will review it soon 

Thanks for posting!~

dbale-altoros (approved):

PR #512, opened 2023-10-20
fix: always let printed reports send to stdout

KalleChen:
# Why do we need this PR?

- If there're Errors in solhint reports, the output will sent to std_err because it exit(1). I think this should be always sent to std_out since there're actually no "errors" in the process, just "Error" against the solhint rules.

This PR is in order to fix [this](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/811#issuecomment-1162582062)

Also, I have a previous [PR](https://github.com/fvictorio/solhint/pull/2) for this, since this repo has not been maintained then.


dbale-altoros:
@KalleChen we really appreciate getting involved
Can you give me further info of what your intentions are with this pr
Take into account this is making several e2e tests fail... so maybe we can work something out with this if I can understand better your point

KalleChen:
> @KalleChen we really appreciate getting involved Can you give me further info of what your intentions are with this pr Take into account this is making several e2e tests fail... so maybe we can work something out with this if I can understand better your point

@dbale-altoros  Thanks for your reply!

The problem is that we use null-ls to print the solhint on nvim. And this [issue](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/811) indicates that the reports output from solhint sometimes sent with `stderr`.

And I found that this function will check if the reposts include `error` then exit the process with 1
```js
function exitWithCode(reports) {
  const errorsCount = reports.reduce((acc, i) => acc + i.errorCount, 0)
  process.exit(errorsCount > 0 ? 1 : 0)
}
```

I also found using process.exit(1) will let the output of console.log sent to `stderr` not `stdout`, and this cause the issue above. 
The problem is if there's only warning in solhint report, it sent it to `stdout`, but if there's one error in report, it will sent it to `stderr`. 

And I think this should not happen, because there's actually no "error" happen in the linting process, just 'error flag' in report. 

Therefore, this PR is trying to remove the exitWithCode function.

If there's something confusing or I'm wrong, pls let me know. thanks again!

btw, I will get some time to investigate why e2e test failed.

dbale-altoros:
I think I understand your point... let me see what we can do... 
We are planning a new version pretty soon... stay tuned...

Join our discord:
https://discord.gg/4TYGq3zpjs

KalleChen:
Okay, I understand. I don't think this is a very urgent matter, so there's no need to rush.

I will be looking forward to seeing the new version of solhint. Thanks for your help and maintain this project.

dbale-altoros:
@KalleChen thanks to you for your contribution 
I will try to squeeze this into next release

KalleChen:
@dbale-altoros thanks! if there's anything I could help you could just let me know

dbale-altoros:
https://github.com/protofire/solhint/pull/554

PR #515, opened 2023-10-24
feat: require the package with fullname

zouguangxian:
- supports require.resolve(package) in extends field

### DIFF
--- a/lib/config/config-file.js
+++ b/lib/config/config-file.js
@@ -1,4 +1,5 @@
 const fs = require('fs')
+const path = require('path')
 const _ = require('lodash')
 const { cosmiconfigSync } = require('cosmiconfig')
 const { ConfigMissingError } = require('../common/errors')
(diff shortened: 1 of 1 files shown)


dbale-altoros:
@zouguangxian thanks a lot for your PR

Can you explain what this PR does and how is this useful ?
Is this testeable ?
Notice it is not passing the github actions flow

zouguangxian:
> @zouguangxian thanks a lot for your PR
> 
> Can you explain what this PR does and how is this useful ? Is this testeable ? Notice it is not passing the github actions flow

In a monorepo, there will be a common practice of putting the shareable configuration in a package, e.g. @organize/solhint-config, then using this package in the solhint.config.js of individual package:
```
  {
    "extends": "@organize/solhint-config",
    "plugins": [],
    "rules": {
    }
  }
```

but it requires the package name should be like 'solhint:recommended' or 'solhint-config-${path}' in the configGetter function, so that '@organize/solhint-config' can not be supported.

this patch will allow the developer to use the package in this way:
```
  {
    "extends": require.resolve("@organize/solhint-config"),
    "plugins": [],
    "rules": {
    }
  }
```

dbale-altoros:
@zouguangxian thanks for the clarification
We need at least for the PR to pass the e2e tests and also don't have errors on it
Please whenever you have time, please correct it, add tests if possible and we can review it

image


Thanks a lot for your contribution

We are planning a new version pretty soon... stay tuned...

Join our discord:
https://discord.gg/4TYGq3zpjs

zouguangxian:
fixed, please review again.

dbale-altoros (approved):

Issue #571, opened 2024-04-09
Feature request: rule for ordering imports topologically and alphabetically


PaulRBerg:
It would be helpful if Solhint offered a rule for ordering imports topologically and alphabetically.

Rationale: this is a non-opinionated code design that lessens the cognitive load on users.

Example of PR where we manually ordered the imports: https://github.com/sablier-labs/v2-periphery/pull/301

I propose the following logic:

- Order by hierarchy of directories first, e.g. `../../` comes before `../`, which comes before `./`, which comes before `./foo`
- Order alphabetically for each file at the same level, e.g. `./bar` comes before `./foo`

dbale-altoros:
thanks @PaulRBerg for posting
I'll check what can be done about this

PaulRBerg:
Can we sponsor you with a bounty to work on this, @dbale-altoros? feel free to share your Ethereum address on [Telegram](https://t.me/PaulRBerg0).

dbale-altoros:
Hi @PaulRBerg i'm on vacations. I'll contact you this friday o next monday

dbale-altoros:
@PaulRBerg 
I was starting to do this
Just to be on the same page
The rule will support these kind of imports

Import parts of a file
- import {IXTokenFactory, IXToken} from '../../token/interfaces/IXTokenFactory.sol';

Import parts and define alias
- import {IXTokenFactory, IXToken as Token} from '../../token/interfaces/IXTokenFactory.sol';

Import the whole file
- import '../../token/interfaces/IXTokenFactory.sol';

Regarding the hierarchy
I will put a sample of an unsorted imports
Would you mind ordering in the way you think it should be

import './ThisIsAVeryLongFileOnPurposeToTestTheFirstPathShorterThanTheLastOnelooooooooooong.sol';
import {Unauthorized, add as func, Point} from "./Foo.sol";
import "https://github.com/owner/repo/blob/branch/path/to/Contract.sol";
import {Initializable} from './openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol';
import '../../../../token/interfaces/FakeContract1.sol';
import {FakeContract3} from '../../../token/interfaces/FakeContract3.sol';
import './../../../../token/interfaces/AFakeContract1.sol';
import './../token/interfaces/IXTokenWrapper.sol';
import {IXTokenFactory, holaquetal} from '../../token/interfaces/IXTokenFactory.sol';
[...]

PR #583, opened 2024-05-23
Exclude abi.encodeX calls from func-named-parameters

0xCLARITY:
Solhint regularly complains about the `named-parameters` rule when all I'm doing is throwing a bunch of data into `abi.encode()` or `abi.encodePacked()`.

Rather than annotating every single `abi.encodeX()` call with a `//solhint-disable` - I think it makes more sense to exclude `abi.encodeX()` calls from this rule at the rule implementation.

Tested locally on the contract I'm currently writing and works as expected.

### DIFF
--- a/lib/rules/naming/func-named-parameters.js
+++ b/lib/rules/naming/func-named-parameters.js
@@ -76,7 +76,7 @@ class FunctionNamedParametersChecker extends BaseChecker {
         if (qtyNamed === 0 && qtyArgs > this.maxUnnamedArguments) {
           this.error(
             node,
-            `Named parameters missing. MIN unnamed argumenst is ${this.maxUnnamedArguments}`
+            `Named parameters missing. MIN unnamed arguments is ${this.maxUnnamedArguments}`
           )
         }
       }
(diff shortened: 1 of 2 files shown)


0xCLARITY:
@dbale-altoros - I pushed up a commit that I think fixes the linting issues of CI, but I'm not sure how to re-trigger the jobs. Any idea?

dbale-altoros:
> @dbale-altoros - I pushed up a commit that I think fixes the linting issues of CI, but I'm not sure how to re-trigger the jobs. Any idea?

I need to do that... already passed all of the CI tests... great job

dbale-altoros:
@0xCLARITY this week there will be a new release with this addition
thanks for the contribution

0xCLARITY:
@dbale-altoros - any ETA on a new release?

dbale-altoros:
@0xCLARITY yes sorry
I was waiting feedback from this issue
https://github.com/protofire/solhint/issues/571
to put both updates together...
pretty confident next week will be a new release

PR #587, opened 2024-07-01
ADD: Import order rule

dbale-altoros:
Requested
https://github.com/protofire/solhint/issues/571


cruzdanilo:
please, make sure it doesn't conflict with foundry's fmt import sorter: https://book.getfoundry.sh/reference/config/formatter?highlight=sort_imports#sort_imports

dbale-altoros:
@cruzdanilo that is a good point... but remember this is a non mandatory rule and will not be part of the recommended pack... so it can be disabled if it present any conflict

PR #607, opened 2024-11-27
chore: fix 404 status URL

wangjingcun:
The previous link has expired and has been replaced with the latest address.


dbale-altoros:
thanks for this
but the file you modified is generated automatically
you can see at the top the legend

I will fix this in a separate pr with other typos

PR #608, opened 2024-12-01
Fix Typographical Errors Update named-parameters-mapping.md

mdqst:
This update addresses a few typographical issues in the documentation of the `named-parameters-mapping` rule.

- **"Default to warn"** has been corrected to **"Defaults to warn"** to ensure proper grammar in the sentence.
- **"Other naming are not necessary"** has been updated to **"Other names are not necessary"** to use the correct plural form of "name."
- **"Other naming are not enforced"** has been revised to **"Other names are not enforced"** for the same reason.

These corrections improve the clarity and readability of the documentation, making it more consistent and easier for users to understand the rule details.

### DIFF
--- a/docs/rules/naming/named-parameters-mapping.md
+++ b/docs/rules/naming/named-parameters-mapping.md
@@ -12,7 +12,7 @@ title:       "named-parameters-mapping | Solhint"
 Solidity v0.8.18 introduced named parameters on the mappings definition.
 
 ## Options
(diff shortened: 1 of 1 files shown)


dbale-altoros:
thanks for this
this pr points out the typos but it is not fixing them... 
the files you modified are generated automatically
you can see at the top the legend

I will fix this in a separate pr with other typos

PR #614, opened 2024-12-20
fix: rename camel case rules to pascal case

dbale-altoros:
Rules named with camel-case were incorrectly named
This pr replaces camel case naming with pascal case naming which is actually what the rule is doing


juanpcapurro:
fwiw the solidity style guide refers to this convention as [CapWords](https://docs.soliditylang.org/en/latest/style-guide.html#naming-conventions) so using the same name might make sanse

PR #615, opened 2024-12-27
Fix broken documentation links 

Vikt0rPavlik:
- Fix broken documentation links due to incorrect paths


dbale-altoros:
Thanks for your suggestion !! @Vikt0rPavlik 
Keep doing it.

In this particular case, at the top of that file it says:
`This is a dynamically generated file. Do not edit manually.` 
So that is not the correct file to fix it.

Either way is already fixed in develop branch.
Happy new year!

PR #635, opened 2025-02-03
Update E2E.yml Configuration

donatik27:
Removed redundant linting step from E2E workflow:
- Deleted "Run linter" step since linting is already performed in another workflow
- Improves CI/CD efficiency by eliminating duplicate checks
- Reduces build time

Changes in .github/workflows/E2E.yml:
- Removed:
  - name: Run linter
    run: npm run lint

Impact:
- Streamlines CI/CD workflow
- Removes duplicate process
- Maintains code quality checks
- Reduces resource usage

### DIFF
--- a/.github/workflows/E2E.yml
+++ b/.github/workflows/E2E.yml
@@ -79,9 +79,6 @@ jobs:
       - name: List directory contents
         run: dir
 
-      - name: Run linter
-        run: npm run lint
-
       - name: Run E2E Tests
         run: cd e2e && npm install && npm test
 


dbale-altoros:
Hey @donatik27 thanks a lot for participating! we really appreciate
I think we don't agree on this removed line
We realize it is already being done in other workflow but since sometimes we run each workflow separate we like to have it run on each...
Maybe if you are up to it, do a separate workflow for the common tasks in the actual workflows and we keep running that new one always, we can remove duplicated tasks (like the one you removed), and can also run each workflow separately after the new one already ran linter... we can discuss this.,.. I just thought this, not sure if worth the hassle
Anyway we appreciate your feedback

PR #637, opened 2025-02-11
Fix link

PixelPil0t1:
#### File: `docs/rules/naming/contract-name-pascalcase.md`

1. **Changed a line**: Updated the link to the rule source.
   - **Old Link**: 
     - [Rule source] (https://github.com/protofire/solhint/blob/master/lib/rules/naming/contract-name-pascalcase.js)
   - **New Link**: 
     - [Rule source] (https://github.com/protofire/solhint/blob/master/lib/rules/naming/contract-name-capwords.js)
   - **Reason**: To correct the reference to the appropriate rule source.

### DIFF
--- a/docs/rules/naming/contract-name-pascalcase.md
+++ b/docs/rules/naming/contract-name-pascalcase.md
@@ -37,6 +37,6 @@ This rule does not have examples.
 This rule is introduced in the latest version.
 
 ## Resources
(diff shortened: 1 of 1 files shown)


PixelPil0t1:
@dbale-altoros

dbale-altoros:
@PixelPil0t1 thanks for the feedback
This is file is generated automatically so that is not the place to change it
Although you made me realize this file is not used anymore, so I delete it in another PR

Thanks!!!

PR #645, opened 2025-02-20
fix: return empty array instead of nothing when formatter is json

dbale-altoros:


lib/formatters/json.js
```diff
@@ -58,5 +58,5 @@ module.exports = function (results) {
 
   if (finalMessage) allMessages.push(finalMessage)
 
-  return allMessages.length > 0 ? JSON.stringify(allMessages) : ''
+  return allMessages.length > 0 ? JSON.stringify(allMessages) : '[]'
```
jirutka, line 61: This ternary is redundant, you can just `return JSON.stringify(allMessages)`.

dbale-altoros: jajajaj yes you are right!

PR #647, opened 2025-03-01
fix: rename camel case rules to pascal case

leopardracer:
Rules named with camel-case were incorrectly named


dbale-altoros:
HEY @leopardracer thanks for the contribution!

PR #650, opened 2025-03-04
fix: grammatical errors and improve wording consistency

Hopium21:
Old: This rule is introduced in the latest version.

New: This rule was introduced in the latest version.

Reason: The correct tense here is past tense ("was introduced") because the rule has already been added in a previous version.

Old: This rule accepts a string option of rule severity. Must be one of "error", "warn", "off". Defaults to warn.

New: This rule accepts a string option for rule severity. Must be one of "error", "warn", or "off". Defaults to warn.

Reason: Changed "of" to "for" for better phrasing, and added "or" for correct list separation.


dbale-altoros:
Hey @Hopium21 thanks a LOT for your contribution. We really appreciate feedback.
Regarding your input. I agree with you on both.

BUT the files you changed are generated automatically so it is of no use your modification. If you want this to be merged please search the process that adds those lines and change that there.

THANKS!!!

PR #653, opened 2025-03-10
feature: replace chalk with picolors

dbale-altoros:
Thanks to [@talentlessguy](https://github.com/talentlessguy)
For this idea
Made some fixes to his original PR


dbale-altoros:
closed because picocolor adds a lot of escape sequences that makes the CI fail
it needs some more work on each formatter

PR #660, opened 2025-03-29
Update contract-name-camelcase.md

zeevick10:


dbale-altoros:
hello @zeevick10 thanks a lot for contributing.
In this particular case, as you can see in line 2 of the file it says
"warning: "This is a dynamically generated file. Do not edit manually."

so that is not the place to do the modification
anyway, that file will be removed in next version

thanks!

PR #674, opened 2025-06-02
fix closing parentheses

otc-png:


dbale-altoros:
@otc-png THANKS a lot for your feedback... keep posting!

Regarding this PR... as you can see at the top of the file

So if you still would like to contribute, please do it in the correct file which is 
`lib/rules/gas-consumption/gas-custom-errors.js`

PR #689, opened 2025-07-02
Feature use natspec

dbale-altoros:
# use-natspec

## Description
Enforces the presence and correctness of NatSpec tags.

## Options
This rule accepts an array of options:

| Index | Description                                                              | Default Value        |
| ----- | ------------------------------------------------------------------------ | -------------------- |
| 0     | Rule severity. Must be one of "error", "warn", "off".                    | warn                 |
| 1     | A JSON object with natspec properties. See EXAMPLE CONFIG section below. | Check EXAMPLE CONFIG |

### Example Config
```json
{
  "rules": {
    "use-natspec": [
      "warn",
      {
        "title": {
          "enabled": true,
          "ignore": {}
        },
        "author": {
          "enabled": true,
          "ignore": {}
        },
        "notice": {
          "enabled": true,
          "ignore": {}
        },
        "param": {
          "enabled": true,
          "ignore": {}
        },
        "return": {
          "enabled": true,
          "ignore": {}
        }
      }
    ]
  }
}
```

### Notes
- If a function or return value has unnamed parameters (e.g. `function foo(uint256)`), the rule only checks the number of `@param` or `@return` tags, not their names.
- If a function or variable has `@inheritdoc`, the rule skips the validation.
- The rule supports both `///` and `/** */` style NatSpec comments.
- If a custom config is provided, it is merged with the default config. Only the overridden fields need to be specified.

#### Contract with valid NatSpec

```solidity

                /// @title Token contract
                /// @author Me
                /// @notice This contract handles tokens
                contract Token {
                  /// @notice Transfers tokens
                  /// @param to The recipient address
                  /// @param amount The amount to transfer
                  /// @return success Whether the transfer succeeded
                  function transfer(address to, uint256 amount) public returns (bool success) {
                    return true;
                  }
                }
              
```

#### You can disable specific tags globally or by type/name using the `ignore` option in the config. For example:

```solidity
{
                "use-natspec": [
                  "warn",
                  {
                    "title": {
                      "enabled": true,
                      "ignore": {
                        "contract": ["MyContract"],
                        "*": ["LegacyContract"]
                      }
                    },
                    "param": { "enabled": false }
                  }
                ]
              }
```

#### The default configuration enables all checks with no ignore rules:

```solidity
{
                "title": { "enabled": true, "ignore": {} },
                "author": { "enabled": true, "ignore": {} },
                "notice": { "enabled": true, "ignore": {} },
                "param": { "enabled": true, "ignore": {} },
                "return": { "enabled": true, "ignore": {} }
              }
```

#### Missing @notice and incorrect @param and @return tags

```solidity

                /// @title Token contract
                contract Token {
                  /// @param wrongParam Not matching actual parameter
                  function transfer(address to) public returns (bool) {
                    return true;
                  }
                }
                
```


fulldecent (approved):
I reviewed by looking at the documentation for this and also browsing through the pool request file changes. Looks great thank you for implementing this and setting the default to warn!

PR #706, opened 2025-08-08
fix: remove `strip-ansi`

talentlessguy:
Replaces `strip-ansi` with built-in `stripVTControlCharacters` which is available since Node 16.11


talentlessguy:
Thanks for the recommendation, will add

dbale-altoros:
@talentlessguy lint is failing... please always run it before you push
there is also a way to test all e2e by doing
`npm run local:e2e`

